PR #13, opened 2025-01-22
Make further use of build-rs where it is included.

### LINKED ISSUE #11: Inconsistent use of build-rs in tutorial
The tutorial includes this line in the Cargo.toml file:

```toml
[build-dependencies]
#...
build-rs = "0.1.2" # This one is optional
#...
```

Then, `build.rs` uses that module in some places, but omits it in other places:

```rust
fn main() {
    build::rerun_if_changed("main.zng"); # Used here
  
    # Not used here, but this could be simplified to `let crate_dir = build::cargo_manifest_dir()` 
    let crate_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());

    # ...
}
```

From the perspective of someone following the tutorial, it would be smoother if build-rs were used consistently, my preference, or not at all.

camio:
Fixes #11

### DIFF
--- a/book/src/tutorial.md
+++ b/book/src/tutorial.md
@@ -415,8 +415,6 @@ zngur = "latest-version"
 Then fill the `build.rs` file:
 
 ```Rust
-use std::{env, path::PathBuf};
-
 use zngur::Zngur;
 
 fn main() {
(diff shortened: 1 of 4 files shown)


HKalbasi:
Thanks!

PR #34, opened 2025-05-05
Add benchmarking infrastructure

HKalbasi:


thesudoer2:
Hello,
The benchmark infrastructure you've added is great for comparing C++ and Rust code. However, the benchmark output only shows the total time spent in each function, without any detailed breakdown.
In my opinion, to benchmark the functions more precisely and identify performance bottlenecks, it would be helpful to add profiling support — such as generating binaries that can be analyzed with tools like perf.

HKalbasi:
@thesudoer2 You can run the added benchmark with perf and similar tools, for example I ran the benchmark using cargo-flamegraph (which is a perf based tool) with this command ` CARGO_PROFILE_BENCH_DEBUG=true cargo flamegraph --bench simple -- --bench`. But the total time is also important, it enables tracking the performance across commits, measuring improvements and regressions.

PR #40, opened 2025-05-18
Add support for Path aliases

Ryex:
- path expansion changed to consider aliases first
- aliases with relative paths expand using the base local to their use not the base local to their definition

also, the addition of a `Process` stage will allow working on things like  
- processing and merging multiple files,
- warning on undefined type reference in the fields and functions
- templates?
```rust
type<T> ::std::option::Option<T> {
    #layout(size = {T.size}, align = {T.align})
}

type<T,E> ::std::result::Result<T, E> {
    #layout(size = {max(T.size, E.size)}, align = {max(T.align, E.align)})
}

type<T> crate::SomeWrapper<T> {
    #layout(size = {T.size + 12}, align = 8)
}

### DIFF
--- a/examples/osmium/main.zng
+++ b/examples/osmium/main.zng
@@ -8,6 +8,8 @@
     #include <osmium/visitor.hpp>
 "
 
+use ::std::string::String as String;
+
 type str {
     wellknown_traits(?Sized);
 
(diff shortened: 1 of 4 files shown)


HKalbasi:
Thanks! Can you use this feature in some example (other than simple and tutorial ones)? I use examples as some kind of end-to-end test.

Ryex:
Added usage of an alias for `::std::string::String` to the `osmium` example

PR #43, opened 2025-06-02
Generate Merged

nickpdemarco:
Adds a `gm` command which takes arbitrarily many `.zng` files and a required `--in` flag which indicates where the generated files should be placed.

This work should be considered a draft for architecture review by @HKalbasi - before spending more time improving error messages and thinking about semantic edge cases, I wanted to see if this passes the smell test.

See `examples/example-simple-merged` to get a sense of how it works. Entities are semantically merged at the parsing stage - unless entities conflict (e.g different layout specs), their fields are merged in predictable ways: usually, that means a union of things like methods, constructors, fields, etc.

### DIFF
--- a/zngur-def/src/merge.rs
+++ b/zngur-def/src/merge.rs
@@ -1,0 +1,137 @@
+use crate::{
+    AdditionalIncludes, ConvertPanicToException, ZngurExternCppFn, ZngurExternCppImpl, ZngurFile,
+    ZngurFn, ZngurTrait, ZngurType,
+};
+use std::vec::Vec;
+
+pub trait Merge<T = Self> {
+    fn merge(self, into: &mut T) -> MergeResult;
+}
+
+pub enum MergeFailure {
+    Conflict(String),
+}
+
+pub type MergeResult = Result<(), MergeFailure>;
+
+fn push_unique<T: Eq>(item: T, smallvec: &mut Vec<T>) {
+    if !smallvec.contains(&item) {
+        smallvec.push(item);
+    }
+}
+
+fn merge_unique<T: Eq>(other: Vec<T>, smallvec: &mut Vec<T>) {
+    for item in other {
+        push_unique(item, smallvec);
+    }
+}
+
+impl<K, V, I: IntoIterator<Item = (K, V)>> Merge<std::collections::HashMap<K, V>> for I
+where
+    K: Eq + std::hash::Hash,
+    V: Merge,
+{
+    fn merge(self, into: &mut std::collections::HashMap<K, V>) -> MergeResult {
+        for (key, value) in self {
+            match into.entry(key) {
+                std::collections::hash_map::Entry::Vacant(e) => {
+                    e.insert(value);
+                }
+                std::collections::hash_map::Entry::Occupied(mut e) => {
+                    match value.merge(e.get_mut()) {
+                        Ok(()) => {}
+                        Err(message) => {
+                            return Err(message);
+                        }
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+}
+
+impl Merge for ZngurType {
+    fn merge(self, into: &mut Self) -> MergeResult {
+        if self.ty != into.ty {
+            panic!(
+                "Attempt to merge different types: {} and {}",
+                self.ty, into.ty
+            );
+        }
+
+        if self.layout != into.layout {
+            return Err(MergeFailure::Conflict("Layout mismatch".to_string()));
+        }
+
+        merge_unique(self.wellknown_traits, &mut into.wellknown_traits);
+        merge_unique(self.methods, &mut into.methods);
+        merge_unique(self.constructors, &mut into.constructors);
+        // TODO: cpp_value, cpp_ref
+
+        Ok(())
+    }
+}
+
+impl Merge for ZngurTrait {
+    fn merge(self, into: &mut Self) -> MergeResult {
+        if self.tr != into.tr {
+            panic!(
+                "Attempt to merge different traits: {} and {}",
+                self.tr, into.tr
+            );
+        }
+
+        merge_unique(self.methods, &mut into.methods);
+
+        Ok(())
+    }
+}
+
+impl Merge<ZngurFile> for ZngurType {
+    fn merge(self, into: &mut ZngurFile) -> MergeResult {
+        [(self.ty.clone(), self)].merge(&mut into.types)
+    }
+}
+
+impl Merge<ZngurFile> for ZngurTrait {
+    fn merge(self, into: &mut ZngurFile) -> MergeResult {
+        [(self.tr.clone(), self)].merge(&mut into.traits)
+    }
+}
+
+impl Merge<ZngurFile> for ZngurFn {
+    fn merge(self, into: &mut ZngurFile) -> MergeResult {
+        push_unique(self, &mut into.funcs);
+        Ok(())
+    }
+}
+
+impl Merge<ZngurFile> for ZngurExternCppFn {
+    fn merge(self, into: &mut ZngurFile) -> MergeResult {
+        push_unique(self, &mut into.extern_cpp_funcs);
+        Ok(())
+    }
+}
+
+impl Merge<ZngurFile> for ZngurExternCppImpl {
+    fn merge(self, into: &mut ZngurFile) -> MergeResult {
+        push_unique(self, &mut into.extern_cpp_impls);
+        Ok(())
+    }
+}
+
+impl Merge<ZngurFile> for AdditionalIncludes {
+    fn merge(self, into: &mut ZngurFile) -> MergeResult {
+        into.additional_includes.0 += self.0.as_str();
+        Ok(())
+    }
+}
+
+impl Merge<ZngurFile> for ConvertPanicToException {
+    fn merge(self, into: &mut ZngurFile) -> MergeResult {
+        // REVISIT: Should we just set to true here?
+        into.convert_panic_to_exception.0 |= self.0;
+        Ok(())
+    }
+}
--- a/zngur-parser/src/lib.rs
+++ b/zngur-parser/src/lib.rs
@@ -5,9 +5,10 @@ use chumsky::prelude::*;
 use itertools::{Either, Itertools};
 
 use zngur_def::{
-    LayoutPolicy, Mutability, PrimitiveRustType, RustPathAndGenerics, RustTrait, RustType,
-    ZngurConstructor, ZngurExternCppFn, ZngurExternCppImpl, ZngurFile, ZngurFn, ZngurMethod,
-    ZngurMethodDetails, ZngurMethodReceiver, ZngurTrait, ZngurType, ZngurWellknownTrait,
+    AdditionalIncludes, ConvertPanicToException, LayoutPolicy, Merge, MergeFailure, Mutability,
+    PrimitiveRustType, RustPathAndGenerics, RustTrait, RustType, ZngurConstructor,
+    ZngurExternCppFn, ZngurExternCppImpl, ZngurFile, ZngurFn, ZngurMethod, ZngurMethodDetails,
+    ZngurMethodReceiver, ZngurTrait, ZngurType, ZngurWellknownTrait,
 };
 
 pub type Span = SimpleSpan<usize>;
@@ -81,19 +82,19 @@ enum ParsedItem<'a> {
         items: Vec<Spanned<ParsedTypeItem<'a>>>,
     },
     Trait {
-        tr: ParsedRustTrait<'a>,
+        tr: Spanned<ParsedRustTrait<'a>>,
         methods: Vec<ParsedMethod<'a>>,
     },
-    Fn(ParsedMethod<'a>),
+    Fn(Spanned<ParsedMethod<'a>>),
     ExternCpp(Vec<ParsedExternCppItem<'a>>),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 enum ParsedExternCppItem<'a> {
-    Function(ParsedMethod<'a>),
+    Function(Spanned<ParsedMethod<'a>>),
     Impl {
         tr: Option<ParsedRustTrait<'a>>,
-        ty: ParsedRustType<'a>,
+        ty: Spanned<ParsedRustType<'a>>,
         methods: Vec<ParsedMethod<'a>>,
     },
 }
@@ -156,6 +157,20 @@ impl ParsedMethod<'_> {
             inputs: self.inputs.into_iter().map(|x| x.to_zngur(base)).collect(),
             output: self.output.to_zngur(base),
         }
+    }
+}
+
+fn checked_merge<T, U>(src: T, dst: &mut U, span: Span)
+where
+    T: Merge<U>,
+{
+    match src.merge(dst) {
+        Ok(()) => {}
+        Err(e) => match e {
+            MergeFailure::Conflict(s) => {
+                create_and_emit_error(&s, span);
+            }
+        },
     }
 }
 
@@ -324,9 +339,9 @@ Use one of `#layout(size = X, align = Y)`, `#heap_allocated` or `#only_by_ref`."
                         ty.span,
                     );
                 };
+                let span = ty.span;
                 let ty = ty.inner.to_zngur(base);
-                r.types.insert(
-                    ty.clone(),
+                checked_merge(
                     ZngurType {
                         ty,
                         layout,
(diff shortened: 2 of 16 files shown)


HKalbasi:
> I added a simple import directive (with pretty terrible error messages) in this PR.

That's great!

> If you'd like to remove gm and just use import, let me know.

I'm happy with removing the `gm` command if you don't need it.

> ZngurSpec really only has semantic entities, no spans, so emitting a rich error is a bit tricky. I'm curious if you have a solution for this.

I guess we will need to pass spans in higher level representations as well. It would also be helpful for late stage errors that we currently don't have, like raising error about missing types.

For testing those errors, we will probably need a function that creates a virtual filesystem like [tests in rust-analyzer](https://github.com/rust-lang/rust-analyzer/blob/cf969d21c30fed7eac32d9e485dca4b39df826f8/crates/ide-assists/src/handlers/move_to_mod_rs.rs#L60) (Using array of file contents instead of comments makes things cleaner and simpler I think, but exact details don't matter).

examples/simple_import/README.md
```diff
@@ -0,0 +1,10 @@
+# Example: Simple Imports
+
+A derivation of the `simple` example to demonstrate the `import` functionality - splits the `simple` example into two zng files which are imported by `main.zng` to demonstrate import semantics.
+
+To run this example:
+
+```
+make
```
camio, line 8: I had to modify this command to make it work on my MacOS machine

```suggestion
make CXX="clang++ -std=c++11"
```

nickpdemarco:
> I'm happy with removing the gm command if you don't need it.

Yeah, I'll pull it to keep the API as simple as possible. 

I'm trying to figure out the semantics of `cpp_value`, `cpp_ref`, and `layout_policy` - it seems that `cpp_ref` implies `!layout_policy`, but I can't find any other currently-coded constraints. As I write the merge logic for these, is there anything else I should keep in mind?

examples/simple_import/Makefile
```diff
@@ -0,0 +1,10 @@
+a.out: main.cpp generated.h src/generated.rs src/lib.rs ../../target/release/libexample_simple_import.a
+	${CXX} -Werror main.cpp generated.cpp -g -L ../../target/release/ -l example_simple_import
+
+../../target/release/libexample_simple_import.a: FORCE
+	cargo build --release
+
+generated.h generated.cpp ./src/generated.rs: main.zng
+	cd ../../zngur-cli && cargo run g ../examples/simple_import/main.zng
+
+FORCE: ;
```
camio, line 10: The `.PHONY` target is the proper way to force rebuilds in makefiles

```suggestion
.PHONY: ../../target/release/libexample_simple_import.a

a.out: main.cpp generated.h src/generated.rs src/lib.rs ../../target/release/libexample_simple_import.a
	${CXX} -Werror main.cpp generated.cpp -g -L ../../target/release/ -l example_simple_import

../../target/release/libexample_simple_import.a:
	cargo build --release

generated.h generated.cpp ./src/generated.rs: main.zng
	cd ../../zngur-cli && cargo run g ../examples/simple_import/main.zng
```

camio (commented):
## Making the example more relatable

To make this example a bit less contrived, you can create a couple C++ header/cpp pairs that make use of different rust entity subsets. You'd then create a corresponding .zng file for each.

```
foo.h
foo.cpp
foo.zng
bar.h
bar.cpp
bar.zng
main.cpp
main.zng
```

```
// main.zng
import "./one.zng";
import "./two.zng";

/* any entities `main.cpp` makes use of */
```

Ideally, "foo" and "bar" would be named after what functionality they each provide.

## Import search path

There isn't a command-line interface to set the import search path, but I think we'll need this to avoid forcing a directory layout with multiple zng-containing dependencies. I think we should allow for multiple `-I` command-line options and, if there are none, default to the directory zng was invoked from. This is what protobuf does.

nickpdemarco:
@HKalbasi @camio I have gone ahead and implemented support for `@dep/path/to/file.zng`. Using it requires passing `--cargo-manifest Cargo.toml` argument (and a `--package FOO` flag if the manifest describes a workspace) to @dep to the path to the crate on disk as downloaded by cargo.

I think this gives us a simple but powerful way to start building out complex projects that contain dependencies which themselves use zngur. I wanted to bring this work to your attention since I intend to raise that follow-up PR if this one is approved & merged.

nickpdemarco:
@Ryex thanks for looking over the code! But let's move discussion of the `@deps` implementation over to [its PR](https://github.com/nickpdemarco/zngur/pull/1). I think the concerns you raise are valid and merit improvements to the design. I just converted that PR to a draft because there's more work to be done to handle crate renames in `Cargo.toml`.

nickpdemarco:
@HKalbasi I believe this is ready for a proper review. As you can see in the parser tests, the error messages are not ideal, but fixing them requires redesigning some parser APIs that would make this (already large) PR unwieldy. 

I'm not quite sure why CI is failing. It passes on my machine, and I can see the file [it's complaining about](https://github.com/HKalbasi/zngur/actions/runs/16276863385/job/45957763931?pr=43) is checked in properly. Any ideas?

examples/simple_import/main.cpp
```diff
@@ -0,0 +1,41 @@
+#include <iostream>
+#include "generated.h"
+
+// External function declarations from foo.cpp and bar.cpp
+extern rust::std::vec::Vec<int32_t> foo();
+extern rust::std::option::Option<rust::std::string::String> bar();
+
+int main() {
+    std::cout << "=== Import/Merge Demonstration ===" << std::endl;
```
skye-tan, line 9: I think this should be removed since the other examples do not contain headings like this.

examples/simple_import/main.cpp
```diff
@@ -0,0 +1,41 @@
...
+    std::cout << "  some_value.is_some(): " << some_value.is_some() << std::endl;
+    std::cout << "  none_value.is_none(): " << none_value.is_none() << std::endl;
+
+    // Use extended API defined in main.zng
+    auto unwrapped_string = some_value.unwrap();  // Use main.zng API
+    std::cout << "  Unwrapped string: ";
+    zngur_dbg(unwrapped_string);
+    std::cout << std::endl;
+
+    std::cout << "=== Extended APIs from main.zng working alongside imported APIs ===" << std::endl;
```
skye-tan, line 39: This should be removed too.

zngur-generator/src/cpp.rs
```diff
@@ -1689,6 +1689,12 @@ namespace rust {
         for ty in [8, 16, 32, 64]
             .into_iter()
             .flat_map(|x| [format!("int{x}_t"), format!("uint{x}_t")])
+            .chain([
+                "long".to_string(),
+                "long long".to_string(),
+                "unsigned long".to_string(),
+                "unsigned long long".to_string(),
```
skye-tan, line 1696: Could you please clarify why this section is needed?
I'm not sure if it is required.

zngur-parser/src/lib.rs
```diff
@@ -455,71 +486,103 @@ Use one of `#layout(size = X, align = Y)`, `#heap_allocated` or `#only_by_ref`."
...
+                match AdditionalIncludes(s.to_owned()).merge(r) {
+                    Ok(()) => {}
+                    Err(_) => {
+                        unreachable!() // For now, additional includes can't have conflicts.
+                    }
+                }
             }
             ProcessedItem::ConvertPanicToException => {
-                r.convert_panic_to_exception = true;
+                match ConvertPanicToException(true).merge(r) {
```
HKalbasi, line 580: Is it possible to restrict some configs (including convert panic to exception) to the top-level file? I guess it may appear surprising if some random imported file make this kind of global change and the top-level file couldn't override it.

zngur-def/src/merge.rs
```diff
@@ -0,0 +1,242 @@
...
+
+        if self.layout != into.layout {
+            return Err(MergeFailure::Conflict(
+                "Duplicate layout policy found".to_string(),
+            ));
+        }
+
+        if self.cpp_ref.is_some() && into.layout != LayoutPolicy::ZERO_SIZED_TYPE {
+            return Err(MergeFailure::Conflict(
+                "cpp_ref implies a zero sized stack allocated type".to_string(),
```
HKalbasi, line 125: This isn't a merge error. Is it added in this PR or was it there before? Is there a better place for checking it?

zngur-def/src/merge.rs
```diff
@@ -0,0 +1,242 @@
...
+            ));
+        }
+
+        self.cpp_value.merge(&mut into.cpp_value)?;
+        self.cpp_ref.merge(&mut into.cpp_ref)?;
+
+        inplace_union(self.wellknown_traits, &mut into.wellknown_traits);
+        inplace_union(self.methods, &mut into.methods);
```
HKalbasi, line 133: I think we should raise error if there is a method with different signatures, though one of the `.zng` files is wrong on its own, so it is not much important.

zngur-def/src/merge.rs
```diff
@@ -0,0 +1,242 @@
...
+        inplace_union(self.constructors, &mut into.constructors);
+        inplace_union(self.fields, &mut into.fields);
```
HKalbasi, line 135: Similar to functions, these also should handle conflict based on name. This is specially problematic for constructor as equivalent constructors can be distinct. For example:
```
Foo { a: i32, b: i64 }
Foo { b: i64, a: i32 }
```

book/src/import.md
```diff
@@ -0,0 +1,87 @@
...
+```zng
+import "path/to/file.zng";
+```
+
+## Path Resolution
+
+Import paths are resolved relative to the directory containing the current `.zng` file:
+
+- `import "./types.zng";` or
+- `import "subdir/types.zng";`
```
HKalbasi, line 16: What do you think about forcing the "./foo/bar.zng" and reserving "foo/bar.zng" similar to js/npm? We can later use `foo/bar.zng` for `bar.zng` file of the `foo` crate.

HKalbasi (approved):
Thanks for this PR, it looks very good! Github says it has conflict with the main branch. Can you please rebase it and squash fix commits?

zngur-generator/src/cpp.rs
nickpdemarco: Thank you for pointing this out - I added this to resolve a build issue in a downstream WASM project of mine (which uses zngur). I've added [a wasm32 example](https://github.com/HKalbasi/zngur/pull/43/files#diff-4a8fc74df49e68d2183fe2b95c0ac341bc7ec562b52a04b390d1b10e40e9d3c5) which fails to compile if we remove this from the codegen path. Without these, we get missing symbol errors at link time. I do not yet have a complete understanding as to _why_, but at least now we have a test to demonstrate it's necessary. Please do take a look and let me know what you think.

I'd be happy to split this into a separate PR if that's preferred.

**Without this change**:

```
make main.js
cargo run --release --manifest-path ../../zngur-cli/Cargo.toml g main32.zng
    Finished `release` profile [optimized] target(s) in 0.03s
     Running `/Users/demarco/repos/metarust/zngur-fork/target/release/zngur g main32.zng`
export RUSTFLAGS="-C link-arg=-sWASM=1 -C link-arg=--no-entry -C link-arg=-sSIDE_MODULE=1 -C link-arg=-fPIC"; \
	cargo build --target=wasm32-unknown-emscripten --release
   Compiling example-tutorial-wasm32 v0.6.0 (/Users/demarco/repos/metarust/zngur-fork/examples/tutorial-wasm32)
    Finished `release` profile [optimized] target(s) in 1.84s
em++ -std=c++14 -sMAIN_MODULE=1 -sWASM=1 -fexceptions -sDISABLE_EXCEPTION_CATCHING=0 main.cpp target/wasm32-unknown-emscripten/release/example_tutorial_wasm32.wasm generated.cpp -o main.js
error: undefined symbol: _ZN4rust25__zngur_internal_data_ptrImEEPhRKT_ (referenced by top-level compiled C/C++ code)
warning: Link with `-s LLD_REPORT_UNDEFINED` to get more information on undefined symbols
warning: To disable errors for undefined symbols use `-s ERROR_ON_UNDEFINED_SYMBOLS=0`
warning: __ZN4rust25__zngur_internal_data_ptrImEEPhRKT_ may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: _ZN4rust30__zngur_internal_assume_deinitImEEvRT_ (referenced by top-level compiled C/C++ code)
warning: __ZN4rust30__zngur_internal_assume_deinitImEEvRT_ may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
Error: Aborting compilation due to previous errors
em++: error: '/Users/demarco/repos/emsdk/node/22.16.0_64bit/bin/node /Users/demarco/repos/emsdk/upstream/emscripten/src/compiler.js /var/folders/zm/ll4504dd6dgbvy20y9ls2_hr0000gq/T/tmp6qk0jeda.txt' failed (returned 1)
make: *** [main.js] Error 1
```

nickpdemarco: Note the demangled names are:

```
_unsigned char* rust::__zngur_internal_data_ptr<unsigned long>(unsigned long const&)
_void rust::__zngur_internal_assume_deinit<unsigned long>(unsigned long&)
```

zngur-parser/src/lib.rs
```diff
@@ -619,27 +682,61 @@ impl ParsedRustPathAndGenerics<'_> {
...
+    path: Option<std::path::PathBuf>,
     text: &'a str,
+    filename: String,
 }
 
 impl<'a> ParseContext<'a> {
-    fn new(filename: &'a str, text: &'a str) -> Self {
-        Self { filename, text }
+    fn new(path: Option<std::path::PathBuf>, text: &'a str) -> Self {
+        let filename = path.as_ref().map_or("<string literal>".to_string(), |p| {
```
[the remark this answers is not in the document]
nickpdemarco: Good question. I made this change because I wasn't sure if "main.zng" was a magic name. As in, does it have any special semantic meaning? It's used in a lot of examples, but as far as I can tell that's convention, not required.

I'm happy to put it back the way it was, because I see what you mean about the tests seeming different/broken now. I wonder if you'd be okay with `test.zng` or something, to nudge folks towards the understanding that `main.zng` isn't a special name.

zngur-parser/src/lib.rs
nickpdemarco: Maybe, but it would be tricky. I agree that this could be surprising, but to add this restriction would be to create two "classes" of zng files: those which can be imported and those which cannot. The latter class would be allowed to `#convert_panic_to_exception` (and perhaps other things), but the former could not. This presents a challenge for library vendors - which class of zngur file are they expected to distribute with their library?

I think the existing functionality is close to correct, but perhaps incomplete. A library vendor should be able to declare, "this crate requires CPTE" (or whatever) and the importing crate must abide by that requirement. But this leaves room for improvement, I agree. Maybe a warning could be emitted, but I'm skeptical that would add much value to the end user.

book/src/import.md
nickpdemarco: Good call. Done!

zngur-def/src/merge.rs
HKalbasi: So it is now checked in two places? Can we defer this check to the post merged file and unify the checks? It is more a semantic error than a merge error, and we may want to add more of these semantic checks in the future, having them in two places makes things a little hard to maintain.

zngur-parser/src/lib.rs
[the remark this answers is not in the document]
HKalbasi: I'm fine with `test.zng`.

nickpdemarco:
@HKalbasi great suggestion. Let me look into that.

zngur-parser/src/lib.rs
nickpdemarco: > This is not great, but I think it is better than a crate become panic=abort if one of its transitive dependencies uses panic=abort

That's fair - I'll look into this.

zngur-def/src/merge.rs
nickpdemarco: I think we should definitely reduce the duplication here. `merge.rs` seems like a pragmatic place to put these types of checks (rather than deferring the check to the post merged file) because here we'll have an easier time referencing back to the .zng files that produced these semantic entities. That means we'll be able to produce error messages that show both of the source code locations that caused the conflict.

Do you think this is required for this PR? I think this might be better handled in a follow-up PR, where I/we can look at improving the parser error messages?

HKalbasi: > Do you think this is required for this PR?

No, I think a `TODO` comment would suffice.

zngur-parser/src/lib.rs
nickpdemarco: Done - CPTE can only be used in top-level files now, and errors otherwise. We have a test confirming this.

nickpdemarco:
@HKalbasi I believe I've addressed all of the current feedback. Ready for your eyes again!

HKalbasi:
Thank you, @nickpdemarco! I think this is ready to merge. Github does not allow me to merge this due to conflicts. Can you please rebase it on main?

nickpdemarco:
Okay, I'm not sure what state we got into there but I've rewritten the branch history off of the merge-base with main.

zngur-parser/src/lib.rs
```diff
@@ -266,8 +288,26 @@ impl ProcessedItem<'_> {
...
+                        "Absolute paths imports are not supported.",
+                        path.span,
+                    )
+                }
+                if !path.path.starts_with(".") && !path.path.starts_with("..") {
+                    create_and_emit_error(
+                        ctx,
+                        "Module import is not supported. Use a relative path instead.",
+                        path.span,
+                    )
```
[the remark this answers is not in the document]
nickpdemarco: True. I'll make the change.

PR #46, opened 2025-06-17
Try to get `cargo test` passing

nickpdemarco:
Cargo test was not run in CI and the tests have regressed.

I have removed the `osmium` example because it could not compile on macOS, but we can put it back before merging if I get some assistance there. Otherwise, the errors appear related to parsing errors.


HKalbasi:
I guess the osmium should work if you install [the library](https://github.com/osmcode/libosmium), but I don't know mac so we can remove it for now and readd it later I think.

nickpdemarco:
@HKalbasi I think I have the tests working now. The main issue was that LATEST_FILENAME and LATEST_TEXT, each wrapped in mutexes, didn't account for the fact that `cargo test` runs all tests asynchronously. So we'd get bogus outputs as errors were reported for files totally unrelated to the failing test. I've added a `ParseContext` to wrap the filename and text and pass it through to the relevant functions.

Now, CI will run `cargo test` as part of `cargo xtask ci`.

HKalbasi:
The current CI failure is probably due some change in the compiler lint (the compiler version is not pinned in CI) that now detects that function as unused. We need to add some `#[allow(unused)]` in the generated code.

PR #49, opened 2025-08-14
Get underling values of Fields

nickpdemarco:
Currently, it does not seem possible to read the underling value of zngur fields in C++. This PR adds support for converting a `Field<T>` to a `T`, and adds corresponding docs and tests.


zngur-generator/src/cpp.rs
```diff
@@ -1658,13 +1658,25 @@ namespace rust {
     struct RefMut;
 
     template<typename T, size_t OFFSET>
-    struct FieldOwned;
+    struct FieldOwned {
+        inline ::rust::Ref<T> as_ref() const noexcept { return ::rust::Ref<T>(*this); }
+        inline T read() const noexcept { auto r = as_ref(); return *r; }
```
HKalbasi, line 1663: I like the `operator T()` but I'm not sure about `.read()` and `.as_ref()`. They may have name collisions with the methods on `T` (specially `.as_ref` which is a method on `Option<T>` and many other types in the std) and they are not needed in the Rust syntax. The user can do `T(some.field)` instead of `some.field.read()` and `Ref<T>(some.field)` for `some.field.as_ref()`, so do these functions pull their weight?

book/src/call_rust_from_cpp/special_types.md
```diff
@@ -17,3 +17,42 @@ tries to support them using C++ feature called User-defined Literals.
 | `'a'_rs_b`     | `b'a'`          | `uint8_t`                         | Not Implemented | unconditionally     |
 | `"hello"_rs_b` | `b"hello"`      | `rust::Ref<rust::Slice<uint8_t>>` | Not Implemented | `[u8]`              |
 | `"hello"_rs_c` | `c"hello"`      | `rust::Ref<rust::ffi::CStr>`      | Not Implemented | `::rust::ffi::CStr` |
+
+## Fields as underlying types
+
+When you declare fields in a tuple or struct using `field name (offset = X, type = T);`, the generated C++ exposes helper wrapper types:
```
HKalbasi, line 23: I would suggest a new section for fields. Fields are not (special) types in the way that `Bool`, `Char`, and `Ref<T>` are types.

examples/regression_test1/main.cpp
```diff
@@ -59,7 +59,34 @@ void test_fields_and_constructor() {
...
+  auto scope = rust::crate::Scoped::new_("Test Field* underlying conversions"_rs);
+
+  rust::Tuple<int32_t, rust::std::string::String> pair{42, "hi"_rs.to_owned()};
+
+  // FieldOwned conversion to Ref and value
+  rust::Ref<int32_t> r0 = pair.f0;
+  int32_t v0 = pair.f0;
+  zngur_dbg(v0);
+  // TODO: Add support for conversion to T for all fields.
+  // rust::std::string::String v1 = pair.f1;
```
nickpdemarco, line 72: @HKalbasi I'm a little stumped here. Still getting my bearings around cpp.rs, is there an obvious place to add an operator T for these?

book/src/call_rust_from_cpp/fields.md
```diff
@@ -0,0 +1,38 @@
...
+When you declare fields in a tuple or struct using `field name (offset = X, type = T);`, the generated C++ exposes helper wrapper types:
+
+- `rust::FieldOwned<T, OFFSET>` for fields on owning types
+- `rust::FieldRef<T, OFFSET>` for fields on `Ref<Ty>`
+- `rust::FieldRefMut<T, OFFSET>` for fields on `RefMut<Ty>`
+
+These wrappers now act as their underlying type `T` in many contexts:
+
+- `Ref<T>` construction from any `Field*<T, OFFSET>`
+- Implicit read via `operator T()` and `.read()` for value-like access
```
HKalbasi, line 12: This `.read()` is leftover.

HKalbasi (approved):

PR #55, opened 2025-09-02
Improve documentation

nickpdemarco:
This PR does a pass over the existing docs for a grammar, spelling, and punctuation correctness.

It also adds `dprint` as a cargo dependency, and a corresponding `cargo xtask format-book` to format our Markdown files sensibly. Feel free to bikeshed any of the settings in `dprint.json`, I have no strong opinions here.

I am a bit opinionated about one difficult-to-enforce policy. In Markdown, I like to try to add line breaks at sentence ends and clause breaks (e.g. at commas, semi-colons, periods, etc) because I find this easier to read in diffs. The best we can do in terms of auto formatting is to use the `maintain` line break setting so clause-breaks are respected by the auto formatter.


HKalbasi:
The CI failes since some crates in dprint lock file are yanked. Are you able to install dprint with `--locked`? It fails in my machine with the same error, but it works without `--locked`.

nickpdemarco:
I am able to install it, but then again I'm on a mac, which seems to be passing in CI. Odd that it's platform-dependent.

HKalbasi:
We can install from binary using their shell script `curl -fsSL https://dprint.dev/install.sh | sh`. It would be faster than installing from source specially for CI which has two cpu cores and we are going to run their code immediately so it doesn't matter security-wise.

HKalbasi:
I ended up using [mise](https://mise.jdx.dev/) to manage dev tools in CI. It manages versions, is cross-platform, doesn't build from source so is fast, and caches installed things. CI is now green, so I will merge this by Monday if you don't have any comment on my commit.

nickpdemarco:
This looks great! Thanks for picking this up, busy week.

PR #58, opened 2025-09-19
AutoZNG implementation using builtin rust tooling

nate-fe2o3:
This implementation targets the direct, automatic creation of the ZngurSpec object from info derived from several rust tools:

- Crate API (types, functions, methods, traits, impl blocks, etc.) generated by the [rustdoc_json](https://docs.rs/rustdoc-json/latest/rustdoc_json/) crate (wrapper around cargo rustdoc) and then deserialized into a Crate object by the [rustdoc_types](https://docs.rs/rustdoc-types/latest/rustdoc_types/) crate.
- Memory Layout info using rustc's `print-type-sizes` nightly feature
- Rust std/core dependency support using the same strategies on statically included json files

The bulk of the work is in the new zngur-def/rdoc.rs file that contains an (incomplete) translation layer from the rustdoc Crate type to the ZngurSpec type. Several types aren't yet supported, but work is ongoing. Standard library dependencies are noted on the first pass of the translation, and then added to the ZngurSpec on separate passes of the RustDocParser.

### DIFF
--- a/zngur-cli/src/main.rs
+++ b/zngur-cli/src/main.rs
@@ -78,11 +78,11 @@ fn main() {
             mangling_base,
             cpp_namespace,
         } => {
-            let pp = path.parent().unwrap();
-            let cpp_file = cpp_file.unwrap_or_else(|| pp.join("generated.cpp"));
-            let h_file = h_file.unwrap_or_else(|| pp.join("generated.h"));
-            let rs_file = rs_file.unwrap_or_else(|| pp.join("src/generated.rs"));
             if auto {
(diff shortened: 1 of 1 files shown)


nate-fe2o3:
I tried my best not to touch any existing code other than adding the `--auto` implementation for the cli. Some notable changes to existing code other than that are the `From<Primitive>` and `From<String>` impl blocks for PrimitiveRustType in zngur-def/lib.rs, as well as marking some types in the examples as public.

HKalbasi:
> where it hits unimplemented functionality in zngur-generator/rust.rs.

I added raw pointers support so we can now continue testing the current implementation.

PR #60, opened 2025-09-29
Fix ci dependency tracking

nickpdemarco:
## Prevent false negatives from stale generated files

The CI system had a critical flaw where `cargo xtask ci` would use stale generated files instead of regenerating them when generator source code changed. Makefiles only tracked `main.zng` changes, so when you modified `cpp.rs` or other sources, `make` would skip regeneration and tests would pass using old generated files while your changes went untested.

The fix adds `clean` targets to all makefiles and modifies CI to run `make clean` before building each example, forcing fresh generation every time. I also added `FORCE` dependencies to generation rules and handled both Make-based and Cargo-based examples properly. Now any changes to the code generation pipeline are immediately tested, eliminating the false positive problem and ensuring local CI behavior matches what actually gets tested.

### DIFF
--- a/examples/memory_management/Makefile
+++ b/examples/memory_management/Makefile
@@ -4,7 +4,10 @@ a.out: main.cpp generated.h src/generated.rs src/lib.rs ../../target/release/lib
 ../../target/release/libexample_memory_management.a: FORCE
 	cargo build --release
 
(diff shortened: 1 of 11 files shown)


nickpdemarco:
Actually, I just cherry-picked this into https://github.com/HKalbasi/zngur/pull/61, so I'll close here.

PR #65, opened 2025-10-28
create unified Scope type for resolving aliases and building paths

Nathan-Fenner:
I'm trying to upstream some changes my team made in a private fork of Zngur (more convenient generic types + some minor C++ interop conveniences). Unfortunately, we fell behind the upstream so it's hard to apply the changes to the current Zngur repo.

Here's a refactor that makes the repo a little more traditionally programming-languagey which should make it easier to refactor/extend in the future.

### DIFF
--- a/zngur-parser/src/lib.rs
+++ b/zngur-parser/src/lib.rs
@@ -56,6 +56,56 @@ struct ParsedPath<'a> {
     span: Span,
 }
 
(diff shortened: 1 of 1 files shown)


HKalbasi:
Thanks for the refactor. Looking forward to your changes!

Issue #48, opened 2025-08-14
Emit separate headers for each sub-crate


nickpdemarco:
With #43 merged, the likelihood of creating massive `generated.h` files increases exponentially. We should have support for splitting `generated.h` into multiple headers.

At first glance, I'd suggest creating a header for each crate that contributed symbols to the `ZngurSpec`. This is easier said than done, since symbols (particularly from std) can be "re-opened" in different zngur files, and we'd expect each zng file that re-opened the symbol to have its FFI code in the crate's generated header.

I think that implies we need to track which .zng files mention which symbols, and then create a header for each _set_ of headers which mention common symbols. That internal, "shared" header would then be included by each individual header to create the desired effect.

```
// a.zng
mod ::std {
    type option::Option<i32> {
        #layout(size = 8, align = 4);

        constructor None;
        constructor Some(i32);
    }
}

type Box<dyn Fn(i32) -> i32> {
    #layout(size = 8, align = 4);
}
```

```
// b.zng
mod ::std {
    type option::Option<i32> {
        #layout(size = 8, align = 4);
        fn unwrap(self) -> i32;
    }

    mod vec {
        type Vec {...}
    }
}
```

```cpp
// internal/generated/ab.h (not exposed to users)
// Insert `option::Option FFI` here
```

```cpp
// generated/a.h (exposed to users)
#include <internal/generated/ab.h>
// Insert `Box<dyn Fn(i32) -> i32>` FFI here
```

```cpp
// generated/b.h (exposed to users)
#include <internal/generated/ab.h>
// Insert `Vec` FFI here
```

HKalbasi:
This is not specific to `Tuple`. `crate1::Type<crate2::Type, crate3::Type>` also has this problem. We can put these types in the dominator of all involved crates, and add an imaginary `final` crate which depends on all crates, ensuring that there is at least one dominator for every set of crates. The problem with this approach is that you may need to include some random crate in order to use a type, but I think it is fine compared to "creating a header for every set of crates" and "manually specifying headers which should contain the items".

PR #67, opened 2025-11-06
Split infrastructure types into zngur.h

nickpdemarco:
Working towards https://github.com/HKalbasi/zngur/issues/48, this PR splits zngur infrastructure types (`rust::Unit`, `rust::Ref`, `rust::RefMut`, and others) into a new `zngur.h` that is `#include`d by `generated.h`.

This obviously comes with some infrastructure changes. Now, clients must be concerned with include paths - notice the changes to the example `Makefile`s to now add `-I.` to their C++ compiler invocations. We also now have a `-o path/to/output/dir` argument to zngur.

I think this is a reasonable milestone to get some feedback on the overall design. Please let me know what you think! If we do merge this, I think it should be a breaking change. We could make the default argument to `-o` be the CWD to approximate backwards compatibility, but I think that runs the risk of confusing users.

### DIFF
--- a/zngur/src/lib.rs
+++ b/zngur/src/lib.rs
@@ -118,7 +118,8 @@ impl Zngur {
             }
 
             // Then add the primary header content, but skip the #include <zngur.h> line
-            let primary_without_include = generated.primary_header
+            let primary_without_include = generated
+                .primary_header
                 .lines()
                 .filter(|line| !line.contains("#include <zngur.h>"))
                 .collect::<Vec<_>>()


HKalbasi:
I don't have a strong opinion on the default. We can even go far and have no default, force user to provide either `--h-file=some_thing.h` or `--h-directory=some_directory/` or some flags like that. We can optimize for the quick demonstrations by using the current single header as the default, but it may reduce the visibility of the multi header feature.

adsnaider:
@nickpdemarco Do you mind if I try to get this over the finish line?

PR #68, opened 2025-11-10
Use explicit calls to `std::ops::Deref::deref`

colin-r-schultz:
Previously, to make method calls through a `deref` we just add `&` and qualify the method name with the `deref` target types. This works in most cases, but fails when the `deref` target is `dyn T`.

```
error[E0277]: the trait bound `&Box<dyn BlobStoreTrait>: BlobStoreTrait` is not satisfied
   --> examples/cxx_demo/src/generated.rs:225:96
    |
225 | ...:<>(&(::std::ptr::read(i0 as *mut &Box<dyn crate::BlobStoreTrait>)), (::std::ptr::read(i1 as *mut u64)), ));
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `BlobStoreTrait` is not implemented for `&Box<dyn BlobStoreTrait>`
    |
    = help: the following other types implement trait `BlobStoreTrait`:
              _zngur_crate_BlobStoreTrait_s12::Wrapper
              _zngur_crate_BlobStoreTrait_s12_borrowed::Wrapper
    = note: required for the cast from `&&Box<dyn BlobStoreTrait>` to `&dyn BlobStoreTrait`
```

This PR changes the generated rust code to use an explicit call to `std::ops::Deref::deref` so that the ambiguity of deref coercion vs `dyn` coercion is resolved and the function compiles.

### DIFF
--- a/examples/simple/expected_output.txt
+++ b/examples/simple/expected_output.txt
@@ -1,7 +1,7 @@
 17
 s[2] = 7
 
-thread panicked at examples/simple/src/generated.rs:192:39:
+thread panicked at examples/simple/src/generated.rs:194:39:
 called `Option::unwrap()` on a `None` value
 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
 s[4] = Rust panic happened


HKalbasi:
You can rebase on the main branch to fix the CI.

PR #77, opened 2025-12-01
collect all errors into the parse context

Ryex:
## Motivation
Currently Zngur emits a parse error and stops parsing immediately once it enters the second parse phase. This means that you'll only see one error at a time as you fix them.

This isn't great both for UX and for adding features that require complex multistage parsing like `#match` or template types.

### Solution

Instead of emitting errors immediately collect reports into the parse context. 
This does mean that once parsing is done we have to check the parse context for errors to ensure the Spec is complete.  

When parsing an import a second `ParseContext` is used and so the context of the parent context consumes it to take it's reports and record it's source cache.

I'm not sure manually checking the context for errors is the best DX, but I'm confident collecting reports is the best path.

### DIFF
--- a/zngur-parser/src/lib.rs
+++ b/zngur-parser/src/lib.rs
@@ -1,4 +1,7 @@
-use std::{fmt::Display, path::Component, process::exit};
+use std::{collections::HashMap, fmt::Display, path::Component};
+
+#[cfg(not(test))]
+use std::process::exit;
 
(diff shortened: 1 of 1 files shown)


Ryex:
> Thanks for the PR! We should avoid adding errors that are follow up to an existing emitted error, but I can't think of any right now. So let's merge this.

the current approach to avoid this is to just not add elements that have caused an error into the spec leaving things incomplete. encountered that when an import statement was in bad form.

PR #78, opened 2025-12-02
`#if` and `#match` statements with `cfg!(key<.value>)` 

Ryex:
Add conditional items with `#if` and `#match` statements that work off pattern matching.

#### Todo
- [x] implement config/feature loading via `CARGO_` environment (build scripts), and via the cli
  - [x] `CARGO_` env for build scripts
  - [x] cli
- [x] Emit items from the body of match arm
- [x] Fix and Add tests
- [x] Document `#if` and `#match` in the book
- [x] Add examples

#### How it works
the backbone are these traits:

```rust
/// a type that can be matched against a Pattern
pub trait Matchable: core::fmt::Debug + Clone + PartialEq + Eq {
    /// A pattern type to match Self against
    type Pattern: MatchPattern;
    /// compare self to `Pattern`
    fn eval(&self, pattern: &Self::Pattern, ctx: &mut ParseContext) -> bool;
}

/// a type that can be matched against a Pattern
pub trait MatchableParse<'src>: Matchable {
    /// return a parser for the item as it would appear in an `#if` or `#match` statement
    fn parser() -> impl ZngParser<'src, Self>;
}

pub trait MatchPattern: core::fmt::Debug + Clone + PartialEq + Eq {}

/// a Pattern tha can be matched against
pub trait MatchPatternParse<'src>: MatchPattern {
    /// return a parser for for the pattern
    fn parser() -> impl ZngParser<'src, Self>;
}

pub trait BodyItem: core::fmt::Debug {
    /// The type Self turns into when added into the Spec
    type Processed;

    /// Transform self into `Processed` type
    fn process(self, ctx: &mut ParseContext) -> Self::Processed;
}

/// a type that hold the body of a conditional statment
pub trait ConditionBody<Pattern: MatchPattern, Item: ConcreteBodyItem>: core::fmt::Debug {
    /// the pattern that guards this body
    fn pattern(&self) -> &Pattern;
}

/// a trait that marks the Cardinality of items inside a body (One? or Many?)
pub trait ConditionBodyCardinality<Item: ConcreteBodyItem>:
    core::fmt::Debug + Clone + PartialEq + Eq
{
    /// the type that hold the body of the conditional statment
    type Body<Pattern: MatchPattern>: ConditionBody<Pattern, Item> + Clone + PartialEq + Eq;
    /// the type that holds the items in the body of a conditional statment
    type Block: core::fmt::Debug + Clone + PartialEq + Eq + IntoIterator<Item = Spanned<Item>>;
    /// the type that hold the items of a passing body
    type EvalResult: IntoIterator<Item = Spanned<<Item as BodyItem>::Processed>>;
    /// transform a single item into Self::Block
    fn single_to_block(item: Spanned<Item>) -> Self::Block;
    /// create a new empty Self::Bock
    fn empty_block() -> Self::Block;
    /// create a new Self::Body from a block and the pattern that guards it
    fn new_body<Pattern: MatchPattern>(
        pattern: Spanned<Pattern>,
        block: Self::Block,
    ) -> Self::Body<Pattern>;
    /// transform a block into it's processed result
    fn pass_block(block: &Self::Block, ctx: &mut ParseContext) -> Self::EvalResult;
    /// trasnform a body into it's processed result
    fn pass_body<Pattern: MatchPattern>(
        body: &Self::Body<Pattern>,
        ctx: &mut ParseContext,
    ) -> Self::EvalResult;
}

/// a trait for a conditional item in a parsed spec
pub trait ConditionalItem<Item: ConcreteBodyItem, Cardinality: ConditionBodyCardinality<Item>> {
    /// Evaluate the statment and produce resulting items of the first arm that passes
    fn eval(&self, ctx: &mut ParseContext) -> Option<Cardinality::EvalResult>;
}

```
The current method for match on a cfg value requires a 

```rust
pub trait RustCfgProvider {
    /// Gets values assoceated with a config key if it's present.
    fn get_cfg(&self, key: &str) -> Option<Vec<String>>;
    /// Gets a list of feature names that are enabeled
    fn get_features(&self) -> Vec<String>;
}

```
 attached to the parse context
 
which can be built both from cargo env variables or from in memory iterables

`#if` and `#match` work analogously to their rust counterparts, passing along the items from the body of the first passing branch.
The parser machinery has a concept of the cardinality of the arm bodies built into the type and the `conditional_item<'a, Item, Condition, Cardinality>(item_parser)` function builds a parser. 

see [the new tests](https://github.com/Ryex/zngur/blob/b7b7b99028717f52b9a7fa4f1c37823a1e851a05/zngur-parser/src/tests.rs#L480) for examples of the syntax.

### DIFF
--- a/book/src/conditional_compilation.md
+++ b/book/src/conditional_compilation.md
@@ -4,6 +4,8 @@ Zngur provides `#if {} #else if {} #else {}` and `#match` statements to control
 which items in you `.zng` file are emitted into the final IDL specification.
 These statements are both powered by pattern matching.
 
+**Note:** These statements are an unstable feature, the syntax may change. You need `#unstable(cfg_if)` and `#unstable(cfg_match)` to enable them.
+
 ## The Scrutinee
 
 The only scrutinee type currently available is `cfg!(key<.item>)`
@@ -49,8 +51,10 @@ To summarize:
 ### Example
 
 ```zng
+#unstable(cfg_if)
+
 // test if "foo" is in the `feature` set
-#if cfg!(feature = "foo") {
+#if cfg!(feature."foo") {
     type crate::Foo {
         #layout(size = 1, align = 1);
     }
@@ -98,8 +102,6 @@ Yow can combine conditions just like in rust
 ```
 
 ## Match Statements
-
-Match statements are an unstable feature, the syntax may change.
 
 ![Match statements](./cfg_match_stmnt.svg)
 
--- a/examples/conditional/main.zng
+++ b/examples/conditional/main.zng
@@ -1,4 +1,5 @@
 #convert_panic_to_exception
+#unstable(cfg_if)
 
 type str {
     wellknown_traits(?Sized, Debug);
--- a/zngur-cli/src/cfg_extractor.rs
+++ b/zngur-cli/src/cfg_extractor.rs
@@ -1,0 +1,225 @@
+use clap::Args;
+use std::collections::HashMap;
+
+#[derive(Args)]
+pub struct CfgFromRustc {
+    /// Load rust cfg values using `rustc --print cfg`
+    ///
+    /// values loaded are in addition to any values provided with `--cfg` or `--feature`
+    /// respects the `RUSTFLAGS` environment variable
+    #[arg(long, group = "rustc")]
+    pub load_cfg_from_rustc: bool,
+
+    /// Use `cargo rustc -- --print cfg` which allows automatically collecting profile flags and
+    /// features
+    ///
+    /// WARNING: because `cargo rustc --print` is unstable this uses `cargo rustc -- --print` which
+    /// may invoke side effects like downloading all crate dependencies
+    ///
+    /// (requires --load-cfg-from-rustc)
+    #[arg(long, group = "cargo_rustc", requires = "rustc")]
+    use_cargo_rustc: bool,
+
+    /// flags to pass to rustc when loading cfg values
+    ///
+    /// (requires --load-cfg-from-rustc)
+    #[arg(long, requires = "rustc")]
+    rustc_flags: Option<String>,
+
+    /// A target to provide to `rustc` when loading config values
+    ///
+    /// (requires --load-cfg-from-rustc)
+    #[arg(long = "target", requires = "rustc")]
+    rustc_target: Option<String>,
+
+    /// cargo profile to use when loading cfg values
+    ///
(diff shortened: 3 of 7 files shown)


Ryex:
@HKalbasi took a bit of work to arrive at a good structure but I think Ive done it. 
parsers and evaluation machinery for `#if` and `#match` statements for any item that can implement `Matchable`.

New tests detailing the syntax are written and pass. let me know if you think there are some edge cases I forgot to cover.
If you think this syntax looks good I can move on to passing config values in through the cli and writing some documentation in the book.

Ryex:
There we go. individual commits should be much easier to review now.

HKalbasi:
Thanks for splitting the commits! It makes reviewing much easier, and I can merge this (relatively huge) PR using rebase instead of squash, retaining its history.

I'm still somehow scare to commit the match syntax, as it has a huge surface with too many tunable parameters, which I don't think the current set of values for those parameters are the best ones. But I don't have an immediate solution to it, and I don't want to block your work. So I propose adding a feature gate statement in the zng file, e.g. `#unstable(cfg_match)`, which enables this match syntax. It doesn't mean much since breaking changes are allowed in Zngur, it is just a signal that this feature is more subject to changes. When we do a Zngur 1.0 release (which I have plan to do in the relatively near future), we can use this infrastructure for adding features without stabilizing them.

About if, I have much less concern, and I think we can enable it without feature gate. Though I like to have `=` inside the `cfg!()` to mimic rust syntax (`cfg!(target_pointer_width = "64")` instead of `cfg!(target_pointer_width) = "64"`), but I see no other tunable options in it. Moving the `=` inside the `cfg!` may make the code reuse between `#if` and `#match` more complex, but I think it worth it.

> The one thing I'm not completely sold on is how I implemented exporting the used CFG values to c++ code.
I'm not sure #define ZNGUR_CFG_<key>_<value> is clean enough. it works for now though.

I think doing it in the preprocessor is the most general place to have this, so it is fine. Do you have an alternative in mind?

HKalbasi:
I made the `#if` unstable as well, to decide about the `cfg!(foo.bar)` syntax, which is not available in Rust later. I also made some minor changes, please tell me if there are problems with my changes. I'm going to merge this now, but I will make follow up PRs if I broke something. Eventually I will revisit the syntax and remove the unstable tags. Thanks for this feature!

Ryex:
no issues on my end.

PR #79, opened 2025-12-02
A few fixes for CI performance

nickpdemarco:
Optimize CI by running zngur binary directly and parallelizing examples
- Change Makefiles to execute pre-built zngur binary instead of cargo run
- Add rayon to parallelize example builds (4 at a time)
- Pre-build zngur-cli once and set ZNGUR env var for all examples
- Reduces CI time by 59% faster on my Apple  M1 laptop

### DIFF
--- a/xtask/src/ci.rs
+++ b/xtask/src/ci.rs
@@ -90,7 +90,6 @@ fn check_examples(sh: &Shell, fix: bool) -> Result<()> {
     sh.change_dir("..");
 
     // Build examples in parallel (limit to 4 at a time to avoid resource exhaustion)
-    // Now each example runs the binary directly - no cargo run overhead!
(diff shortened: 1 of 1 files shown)


HKalbasi:
Nice! This reduces the CI time on github runners for macOS from 6m to 3m. But it doesn't change the time for linux targets, maybe because they are single core on github? I'm not sure.

> Add rayon to parallelize example builds (4 at a time)

I don't think it is correct to use `with_max_iter_len` for limiting build jobs to 4. I guess it will limit the number of parallel jobs to `number_of_examples/4`. What will happen if we remove this constrain entirely? Will it become slower? We can also use jobserver to smartly limit the number of inner tasks to the number of cores, but it might be overkill.

PR #87, opened 2026-01-02
Support zero-sized types

colin-r-schultz:
Switch uses of `&data[0]` to `data.data()` because calling `operator[]` on a `std::array<T, 0>` is always an error, even when we are just getting an address. This prevented the use of zero-size types.

Noting that another trivial fix for this is to define `data` as `std::array<uint8_t, max(size, 1)>` as is done for `()` but I'm not sure there is any advantage there.

### DIFF
--- a/examples/regression_test1/expected_output.txt
+++ b/examples/regression_test1/expected_output.txt
@@ -107,5 +107,6 @@ Test Ref<Ref<T>> -- finished
 
 Test zero-sized type -- started
 Method call on ZST
-[main.cpp:127] zst = ZeroSizedType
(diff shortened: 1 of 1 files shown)


HKalbasi:
This is now better, and those specific branches are fine. I think we can merge this PR after resolving the conflicts and fixing CI.

colin-r-schultz:
From my view I see CI passing and no conflicts

HKalbasi:
That's.... interesting.

image

colin-r-schultz:
Might be some merge vs rebase issue. I tried rebasing my changes on main so hopefully that works.

PR #88, opened 2026-01-08
Template type definitions

colin-r-schultz:
Still a little code cleanup to do, but want to get review on overall design.

Adds the ability to create templated type definitions to `.zng` files, such as the following
```
type<T> ::std::vec::Vec<T> {
    #layout(size = 24, align = 8);
    fn new() -> :std::vec::Vec<T>;
    fn push(&mut self, T);
    fn len(&self) -> usize;

    fn get(&self, usize) -> :std::option::Option<&T> deref [T];
}
```
Now, when a definition like the following is added, the method, trait, and layout information is automatically copied from the template.
```
type ::std::vec::Vec<crate::MyType> {}
```
As shown, this allows for very compact definitions of common generic types.

With the current design, you could not call `Vec<MyType>::get(...)` because `[MyType]` and `Option<&MyType>` are not defined. Template types do not add any new type definitions nor do they force you to add definitions that you aren't going to use. See examples/template_types/main.zng for more details.

Definitely open to feedback on the design. My team has been using a system like this for a while and have found it very convenient.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ members = [
 # Exclude tutorial-wasm32 from default workspace builds since it requires extra dependencies
 exclude = [
     "examples/tutorial-wasm32",
+    "examples/multiple_zngur_files",
 ]
 resolver = "2"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,6 @@ members = [
 # Exclude tutorial-wasm32 from default workspace builds since it requires extra dependencies
 exclude = [
     "examples/tutorial-wasm32",
-    "examples/multiple_zngur_files",
 ]
 resolver = "2"
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -577,6 +577,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "example-template-types"
+version = "0.7.0"
+
+[[package]]
 name = "example-tutorial"
 version = "0.10.0"
 
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -11,6 +11,7 @@
   - [Types with special support](./call_rust_from_cpp/special_types.md)
   - [Panic and exceptions](./call_rust_from_cpp/panic_and_exceptions.md)
   - [Raw pointers](./call_rust_from_cpp/raw_pointers.md)
+  - [Template types](./template_types.md)
 - [Calling C++ from Rust](./call_cpp_from_rust/index.md)
   - [Calling C++ free functions](./call_cpp_from_rust/function.md)
   - [Writing `impl` blocks for Rust types in C++](./call_cpp_from_rust/rust_impl.md)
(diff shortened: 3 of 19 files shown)


Ryex:
very nice. I was working on some thing similar a few days ago but my attempt was using a context parser to identify type vars at parser time instead of later. I think I prefer this approach.

colin-r-schultz:
@hkalbasi Thanks for the review. Sorry this fell off my radar for a while. I've fixed the error reporting for missing `#layout` definitions, and added a few tests. I'm happy to contribute a book chapter as well.

Let me know if there are any other changes you want to see in this pr.

hkalbasi:
Thanks! Will look at it again tomorrow. A book chapter would be very helpful.

adsnaider:
What happens in the common case where the layout of a type is dependent on T @colin-r-schultz?

hkalbasi:
@adsnaider I think the idea is that you put the layout information in the "specialized" version of the type, which you need anyway,

Cargo.toml
```diff
@@ -14,6 +14,7 @@ members = [
 # Exclude tutorial-wasm32 from default workspace builds since it requires extra dependencies
 exclude = [
     "examples/tutorial-wasm32",
+    "examples/multiple_zngur_files",
```
hkalbasi: Why we excluded this? Did you encounter any problems with this example?

colin-r-schultz: That was a mistake. It looks like that example was deleted or renamed but it was still hanging around in my worktree causing cargo errors.

hkalbasi:
Some comments after experimenting with this:
1. I have bitten by methods who disappear because their types are not available, I think it would specially surprise newcomers, so I think at least we need a warning for this. What do you think about disallowing this completely, and requiring all types mentioned in templates to be defined? Do we lose much functionality and flexibility this way? I think I would prefer making this an error if there isn't a huge downside, since we can remove that error in the future without breaking change (but not vice versa). 
2. `Self` is not substituted correctly, leading to strange disappearing of `fn new() -> Self` in vector.
3. This feature has many corner cases (like the `Self` case above), so please add a `template_regression_test` example or something like that which exercise these corner cases deeply, to ensure the current code works as expected and prevent future regressions.

colin-r-schultz:
@hkalbasi I have removed the logic to filter methods based on what types are defined. I agree this makes the usage more intuitive, at the cost of forcing more type definitions. The forward compatibility argument is very fair.

I did not intend to add any support for `Self`. In particular, `Self` is not supported in regular type definitions as far as I can tell and I did not add it as a special case for templates. By getting rid of the filtering logic, you will get a reasonable error message if you try to use `Self`.

I intended for `examples/template_types` to be the regression test suite for this feature as it covers all the common functionality. I'm happy to add more tests if there are other behaviors you think I should cover.

I've also added a book chapter, happy to take any feedback on that.

hkalbasi:
Thank you very much!

PR #90, opened 2026-01-11
Layout conservative and auto offset fields

### LINKED ISSUE #1: Implement `layout_conservative(size = X, align = Y)` layout policy

Ryex:
fixes #1 
- feat: field auto offsets
    - Rework `FieldOwned` and `FieldRef*` types to encode offset as types rather than static `size_t`
    - Rework fields to be designated in `union` with the struct data member so they transparently adopt the address of the   
      allocated data / pointer. This makes fields work in c++11 or greater.
    - Optimize `Ref*<T>` creation for the static case
    - Make Fields work for `#heap_allocated` types by detecting if the parent type of the "first" offset is `heap_allocated`.
    - The data member is now public as a consequence of being in union with the fields so it's been 
      renamed `__zngur_data` to avoid name conflicts and flag it as an implementation detail

- feat: add `#layout-conservative`
- fix: make `Ref*<T>::operator*()` work in const qualified contexts
- fix: ensure generated constructors do default construction first for heap allocation
- feat: add regression test for fields and auto field offset
- feat: add `#layout-conservative` regression test
- chore: ~~run cargo xtask format-templates~~ reverted, breaks things
- Update book for `#layout_conservative` and auto offset fields

### DIFF
--- a/book/src/call_rust_from_cpp/fields.md
+++ b/book/src/call_rust_from_cpp/fields.md
@@ -37,3 +37,9 @@ mt.f1.push_str("!"_rs);
 ```
 
 See `examples/regression_test1` for a runnable demonstration.
+
+## `offset = auto`
+
+If you do not know the internal layout of your type, i.e. when using `#heap_allocated` or `#layout_conservative`,
+you can set the offset to `auto`. This will emit a unique symbol from the rust side with the real offset which will be obtained by c++ at link time via a `extern const size_t`.
+This does have some small performance penalty when using deeply nested fields as the real offset must then be computed at run time for every conversion to a `Ref<T>`.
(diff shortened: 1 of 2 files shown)


HKalbasi:
Hi! Sorry for the delay. I was in an internet blackout. I still don't have a stable connection, will look at your PRs in the following days, hopefully.

Since there is a probability of more such problems in the near future, I want to give commit access to you, as a backup maintainer, in cases where I was not available. Even when I am present, it would be helpful for you to review and merge other PRs. If anything slips through, we can always fix it with a follow-up PR.

What do you think?

Ryex:
> Hi! Sorry for the delay. I was in an internet blackout. I still don't have a stable connection, will look at your PRs in the following days, hopefully.
> 
> Since there is a probability of more such problems in the near future, I want to give commit access to you, as a backup maintainer, in cases where I was not available. Even when I am present, it would be helpful for you to review and merge other PRs. If anything slips through, we can always fix it with a follow-up PR.
> 
> What do you think?

I'm fine with being a backup maintainer. Hopefully things stabilize for you folks.

Issue #47, opened 2025-08-13
`char` not generated properly


asoffer:
Running zngur on this specification

```
type MyType {
  #layout(size = 16, align = 16);
  fn f(char);
}
```

yields output that refers to `::char` which is not valid Rust.

asoffer:
Digging a bit deeper, it looks like the docs mention this but it hasn't been implemented. If I were to take this on,
1. Would a PR be welcome?
2. Is there anything important I should know before starting?

HKalbasi:
> Would a PR be welcome?

Definitely!

> Is there anything important I should know before starting?

We probably also need a convenient way to create `rust::Char` in C++. The `'a'_rs` would work for the ascii case, but I'm not sure about the unicode case. Not a blocker though.

HKalbasi:
Fixed in #96

PR #96, opened 2026-02-20
char generation

mtimaN:
Hi. This PR implements char generation as mentioned in #47.

I treated the Char case very similarly to `rust::Bool` and tested using a small example:
```
// main.zng
type char {
  #layout(size = 4, align = 4);
  wellknown_traits(Copy);
}

type crate::Example {
  #layout(size = 0, align = 1);
  fn g(char);
}
```
```
// main.cpp
#include "./generated.h"

using ::operator""_rc;

int main() {
    char32_t c = 0x000055u;
    auto example = rust::crate::Example::g(operator""_rc(c));
}
```

```
// lib.rs
mod generated;

struct Example {}

impl Example {
    fn g(c: char) {
        println!("printing a char: {}", c);
    }
}
```

Please let me know whether there is anything I missed.

### DIFF
--- a/examples/char/src/lib.rs
+++ b/examples/char/src/lib.rs
@@ -1,17 +1,18 @@
+#[rustfmt::skip]
 mod generated;
 
-pub(crate) struct CharPrinter;
+struct CharPrinter;
 
 impl CharPrinter {
-    pub(crate) fn print(c: char) {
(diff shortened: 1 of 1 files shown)


HKalbasi:
Thanks!

PR #118, opened 2026-04-21
Add C++ stack-owned in Rust for trivially relocatable types

adsnaider:
* Adds support for C++ trivially relocatable values in the Rust stack
  * A couple of examples using this feature
* Added some helper functions to C++ generated pointers and reference types
* Added `align_of` to generated C++ headers
* Fixed size and alignment of reference types to match uintptr_t (instead of assuming 64 bit arch)
* Created a zngur-lib (znglib) crate that is used for trait definitions that C++ autogenerated structs implement
  * This enables generic programming over C++ types (think default constructible, uninitialized, destruction, etc.)
* Add a book entry on trivial relocatable C++ objects


examples/stack_owned/impls.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include "generated.h"
+
+namespace rust {
+namespace exported_functions {
+
+rust::crate::MyCppWrapper create_cpp_type(int32_t x, int32_t y) {
+  rust::crate::MyCppWrapper w;
+  new (&w.cpp()) CppType(x, y);
+  ::rust::__zngur_internal_assume_init(w);
```
hkalbasi: How about generating a constructor on `MyCppWrapper` which takes a `CppType`? `__zngur_internal_assume_init` is clearly an internal zngur function, and we should not advertise using it in the examples.

examples/stack_owned/cpp_type.h
```diff
@@ -0,0 +1,20 @@
...
+  CppType(int x, int y) : x(x), y(y) {
+    std::cout << "Constructed CppType " << x << " " << y << std::endl;
+  }
+  ~CppType() {
+    std::cout << "Destructed CppType " << x << " " << y << std::endl;
+  }
+};
+
+namespace rust {
+template <> struct is_trivially_relocatable<CppType> : std::true_type {};
```
hkalbasi: Do we have a specialization for `is_trivially_copyable`? Let's also add a non `is_trivially_copyable` type (e.g. one that have a std::vector in fields) to this example.

zngur-generator/templates/zng_header.sptl
```diff
@@ -294,6 +301,45 @@ namespace {{ self.cpp_namespace }} {
...
+      }
+
+      template<typename U = T, typename ::std::enable_if<!zngur_is_unsized<U>::value, bool>::type = true>
+      Ref<T> AsRef() const {
+          return Ref<T>::FromAddr(Addr());
+      }
+
+      template<typename U = T, typename ::std::enable_if<zngur_is_unsized<U>::value, bool>::type = true>
+      Ref<T> AsRef() const {
+          return Ref<T>::FromAddrAndMetadata(Addr(), Metadata());
```
hkalbasi: Why we need these functions? `Ref` and `Raw` have constructors which do the job of `AsRef` and `AsRaw`, and the rest of functions have alternatives in Rust std that can be bridged.

book/src/call_cpp_from_rust/trivially_relocatable.md
```diff
@@ -0,0 +1,136 @@
...
+Rust's move semantics can be summarized as "all objects in Rust are trivially
+relocatable", and for the purpose of `zngur` interop, a C++ object is considered
+trivially relocatable if performing a trivial relocatable move on the object
+has the same observable effect as move constructing the C++ object in the new
+memory location and running the moved-from objects destructor.
+
+A lot of C++ objects likely follow this guarantee. A few examples likely include:
+
+- Plain old data types
+- any `T` with a `T(T&&) = default;` implementation
```
hkalbasi: And without any non trivially relocatable members.

book/src/call_cpp_from_rust/trivially_relocatable.md
```diff
@@ -0,0 +1,136 @@
...
+- `std::vector<T>` if `T` is trivially relocatable
```
hkalbasi: Even if `T` is not trivially relocatable?

book/src/call_cpp_from_rust/trivially_relocatable.md
```diff
@@ -0,0 +1,136 @@
...
+  - Most other collections
+- `std::shared_ptr`
+- `std::future`
```
hkalbasi: Also add some negative examples:

- `std::string`
- self-referential data types
- classes with non trivially relocatable members

book/src/call_cpp_from_rust/trivially_relocatable.md
```diff
@@ -0,0 +1,136 @@
...
+
+Furthermore, C++ allows you to `= delete` a move constructor which effectively
+prohibits a value from moving in memory once constructed
+
+## Rust's Move Semantics
+
+Meanwhile, Rust's move semantics are very simple. When you move a value, the
+Rust compiler will effectively copy the struct's bits and invalidate the moved
+from object. The Rust compiler will not run the `Drop` implementation on the
+moved-from value.
```
hkalbasi: I think we don't need that amount of details on C++ value kinds. A simple "But C++ allows running arbitrary code in its move constructor" in the second section would do the job.

book/src/call_cpp_from_rust/trivially_relocatable.md
```diff
@@ -0,0 +1,136 @@
...
+    fn create_cpp_type(i32, i32) -> crate::MyCppWrapper;
+    fn print_cpp_type(&crate::MyCppWrapper);
+}
+```
+
+The C++ code has access to the `.cpp()` method used to access the inner type.
+
+`zngur` will generate stack-owned object definitions inside of a `mod cpp {}`.
+This is different from reference-based objects since we need new definitions for
+each object given the layout differences.
```
hkalbasi: Let's make reference based types use this as well.

examples/cpp_inheritance/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "example-cpp-inheritance"
+version = "0.1.0"
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+publish = false
+
+[dependencies]
+cpp_inherit = { path = "cpp_inherit" }
+znglib = { path = "../../zngur-lib" }
```
hkalbasi: Let's make this `cpp_inherit_complex` and also add a `cpp_inherit_simple` which uses hard coded inheritance and is not generic. I don't want people to see this as the only option on inheritance.

zngur-generator/src/template.rs
```diff
@@ -477,6 +477,31 @@ impl<'a> CppHeaderTemplate<'a> {
...
+"#,
+                                namespace = self.namespace,
+                                ty = imp.ty,
+                                cpp_type = cpp_stack_owned.cpp_type
+                            ));
+                            continue;
+                        }
+                    }
+                }
+
```
hkalbasi: Use a dedicated syntax (per the discussion in the telegram group) for constructors in both stack owned and heap owned types, or if you consider it out of scope for this PR, it's fine, just remove this and we can use manual `new` impls until we have this.

zngur-generator/templates/zng_header.sptl
```diff
@@ -57,7 +57,7 @@
 
 namespace {{ self.cpp_namespace }} {
 
-  inline void* __zng_memcpy( void* dest, const void* src, std::size_t count ){
+  inline void* __zng_memcpy( void* dest, const void* src, ::std::size_t count ){
```
hkalbasi, line 60: Is this change fixing a bug, or is it just style? If it's a bug please add something about it in the regression example.

zngur-lib/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+///
+/// # Safety
+///
+/// Only generated objects by zngur should implement this trait.
+pub unsafe trait ZngCppStackObject: ZngCppObject + Sized {
+    /// Default constructs the C++ stack-allocated object using the the C++ default constructor
+    // NOTE: This is defined in `ZngCppStackObject` because I foresee expanding
+    // initialization to heap allocated or even reference-based opaque objects.
+    // This `new()` function implementation only makes sense for stack allocated
+    // objects.
```
hkalbasi: Let's do that expansion in this PR.

HKalbasi (commented):
Thanks for the PR! I think we should unify `#cpp_stack_owned` and `#cpp_heap_owned` in this PR, because there is a possibility of huge code reuse between them, and some things in `#cpp_heap_owned` may things on the `#cpp_stack_owned` design (specially on the construction/deconstruction).

examples/stack_owned/impls.cpp
adsnaider: I don't have support for non-default constructors and it's no easy task either. For what it's worth, we probably want to support complex construction semantics and I don't want everything to be hidden behind autogenerated magic, so how about we rename/reexport this method as `zngur_assume_init()` for C++ code to use

examples/stack_owned/cpp_type.h
adsnaider: I'm not doing anything with `is_trivially_copyable`. I don't think it's relevant to this PR but we can discuss future improvements

zngur-generator/templates/zng_header.sptl
adsnaider: I didn't see the constructors originally. There's no way to construct a Ref[Mut] from a raw pointer though. Personally I prefer having the `AsRef`, `AsMut`, `AsRaw`, `AsRawMut`, etc. The other methods are just generally useful to get regular C++-style pointers and also to store the fat pointer metadata in the case of inheritance over a trait object for instance (which I'm already doing downstream)

book/src/call_cpp_from_rust/trivially_relocatable.md
adsnaider: Ah yeah, you're totally right. Growing the vector might call move constructors but moving the vector won't...

zngur-generator/templates/zng_header.sptl
adsnaider: I had other places where the missing `::` was causing issues when it conflates with Rust's `std`, so I just changed all occurrences in the generated code. I don't know if this one specifically was breaking but there was an example that was breaking

zngur-lib/src/lib.rs
adsnaider: That's a full breaking change since all downstream crates will need to start importing `znglib`. Not saying we shouldn't do it, but do you really want to make this change a breaking change as well?

book/src/call_cpp_from_rust/trivially_relocatable.md
adsnaider: I agree, we should do this, but I'm not sure we want to do it here. (See my other comment)

examples/stack_owned/impls.cpp
hkalbasi: > I don't have support for non-default constructors and it's no easy task either.

I think this is not related. What I want here is a constructor for the C++ generated type from the real C++ type, for all stack owned (and heap owned) types, regardless of how many constructors the original C++ type has.

> we probably want to support complex construction semantics 

I don't think we would lose any expressivity power, complex construction scenarios would become a move (which its job is just setting the flag).

examples/stack_owned/cpp_type.h
hkalbasi: I just mean a specialization so that user don't need manual specialization of `is_trivially_relocatable` for those types. It is not about `#[derive(Copy)]` and possible similar features.

zngur-lib/src/lib.rs
hkalbasi: We should eventually break this, the current `ZngurOpaqueObject` is too bad to keep. So, if we break it sooner, less code depends on it. And this PR is adding something very related, so I think it's the time.

examples/stack_owned/impls.cpp
adsnaider: But the type doesn't have a default constructor. I could add one to this example but that's not something I want to force on every type since not everything in C++ is default-constructible

hkalbasi: It doesn't need to be default constructible. I mean:

```
class rust::crate::MyCppWrapper {
    MyCppWrapper(); // Default constructor which is equivalent to uninit Rust type with drop_flag=false, and is available on every zngur generated type, keep it.
    MyCppWrapper(CppType&&) // Add this for every stack_owned and heap_owned type. It doesn't need CppType to be default constructible.
}
```

adsnaider: I've gone a slightly different approach by having a `::build()` method that  takes any forward arguments into the constructor so it works for constructing the object with any parameters, including copy and move constructor. That's also consistent with opaque objects. I still think we want to offer the C++ code the option to muck with the drop flag if needs to. Like if it needs to "forget" an object like in Rust you can assume_deinit, and if you want to rebuild the object in place, you can assume_init.

examples/stack_owned/cpp_type.h
adsnaider: Done

book/src/call_cpp_from_rust/trivially_relocatable.md
adsnaider: Done

book/src/call_cpp_from_rust/trivially_relocatable.md
adsnaider: Done

examples/stack_owned/impls.cpp
hkalbasi: Forwarding arguments make sense to me. Why a `build` method instead of using the constructor?

> That's also consistent with opaque objects

We are going to remove them so consistency with them doesn't matter?

> Like if it needs to "forget" an object like in Rust

There are options, e.g. for this, you can pass the object to `rust::std::mem::forget`. I think there are ways for other scenarios, but if you found a scenario without a solution, we can discuss about adding such ways.

zngur-generator/templates/zng_header.sptl
hkalbasi: I generally like to not have additional methods on the Rust mirror objects, in order to not shadowing the potential Rust methods with same name, and to prevent confusion about the origin of the generated methods when someone is reading zngur related code. So I would suggest using the `Raw` and `Ref` constructors (which mimics `&mut foo` and `&raw mut foo` syntax in Rust, respectively) and having an `addr` method (lower case) only on the `Raw` and `RawMut` type, because it [exists in Rust](https://doc.rust-lang.org/std/primitive.pointer.html#method.addr) so it has no ambiguity.

examples/stack_owned/impls.cpp
adsnaider: I still think we want to have the option to have uninitialized types. If I want/need to initialize the C++ object in-place, then I might want to construct the object in an uninitialized state, and then do whatever logic I need to fix it into an initialized state.

mem::forget() won't work if the object has to be pinned

zngur-generator/templates/zng_header.sptl
adsnaider: Done. I also fixed the syntax for read_* since the correct method is as_[ref|mut]_unchecked in the Rust end.

One note on usability. Using the constructors requires you to spell out the internal type completely since type inference in C++  sucks, so I would still like for you to consider adding a convenience function for conversion. Maybe it can be a static function so you would call it like `Ref::as_ptr()`

examples/cpp_inheritance/Cargo.toml
adsnaider: Honestly, this is probably still not the right abstraction, and I'm working on something else upstream, so I'm happy to either remove this completely or keep it as a somewhat crappy POC

zngur-lib/src/lib.rs
adsnaider: These are gone now, replaced by new-type wrappers, but I'm not ready to generalize them as I explained in telegram

examples/stack_owned/impls.cpp
adsnaider: @HKalbasi, it's a `build` method which also matches the allocated objects API. I don't see why change the semantics here. For Rust objects in C++, when you call the default constructor, it's uninitialized. This should be the same, especially since otherwise, there's no way to set it up for in-place construction

hkalbasi: There is a conflict chance with a Rust `build` method, but I guess it's fine.

> mem::forget() won't work if the object has to be pinned

But ManuallyDrop and MaybeUninit would still work. It needs to be discussed on real world scenarios I think.

zngur-generator/templates/zng_header.sptl
hkalbasi: I see your point on type inference, but at least it is unrelated to this PR, so we can discuss it later.

examples/cpp_inheritance/Cargo.toml
hkalbasi: Then I would prefer removing it. Examples are somehow a recommendation on the style, and I don't want to recommend this. You can use the regression examples if you want to prevent regression on code added in this PR which are not covered by other examples, if you want.

book/src/call_cpp_from_rust/opaque.md
```diff
@@ -127,19 +137,153 @@ Note that you don't need the `trait` declaration in the zng file if you just nee
...
-- `std::mem::swap::<RustType>(a, b)` only swaps the first zero bytes of those, i.e. does nothing.
+At construction with `::build` in C++, `zngur` allocates an object of the
+correct size and a alignment, and initializes it with the forwarded arguments.
+The only thing you can do with that object is call C++ methods or `Drop` it
+which calls the destructor and frees up the allocation
 
-Those problem might be solved by the `extern type` language feature.
-Using that, we can define the `ZngurCppOpaqueBorrowedObject` as a proper extern type
-instead of an imaginary zero sized type at the first of C++ objects.
+In the future, we may remove the `destructor` field when its statically known
```
hkalbasi: I think a note that we may replace this with unique_ptr of `#cpp_ref` types, or something like that, is helpful.

book/src/call_cpp_from_rust/opaque.md
```diff
@@ -127,19 +137,153 @@ Note that you don't need the `trait` declaration in the zng file if you just nee
...
+(you can follow along the full example in `examples/stack_owned/`)
+
+```zng
+// main.zng
+#cpp_additional_includes "
+#include <cpp_type.h>
+"
+
+type crate::MyCppWrapper {
+    #cpp_stack_owned "::CppType" (size = 8, align = 4);
```
hkalbasi, line 225: Now that we are going to remove `#cpp_value`, we will end up with `#cpp_ref` and `#cpp_stack_owned`. What about making it `#cpp_owned`?

examples/stack_owned/src/main.rs
```diff
@@ -0,0 +1,22 @@
...
+use znglib::{ZngCppDefaultConstruct, ZngCppStackObject};
+
+// SAFETY: Constructor calls the actual C++ constructor. Object is initialized
+// after calling `.construct()`
+unsafe impl ZngCppDefaultConstruct for MyCppWrapper {
+    unsafe fn construct(&mut self) {
+        self.constructor();
+    }
+}
+
```
hkalbasi: I think this example is clearer without this trait and by adding a manual new (which uses `::build()`) instead of the current construct.

book/src/call_cpp_from_rust/opaque.md
adsnaider: I agree we need to work out the names again, but not yet. The space needs to be explored further. I think a better solution is to indiciate to zngur the restrictions and knowns of a C++ type, which would allow us to generate different semantics for it.

For instance 

```
type crate::MyCppWrapper {
     #cpp "::CppType"
     #layout (size = 8, align = 4)
     #trivially_relocatable
}
```

This tells zngur, that it's a C++ wrapper over "::CppType", with known layout and is trivially relocatable. This means we can use it in the Rust stack, heap, or over references and pointers

```
type crate::MyCppWrapper {
     #cpp "::CppType"
     #trivially_relocatable
}
```

This just says that the type can only be used through a pointer that can be unpinned

```
type crate::MyCppWrapper {
     #cpp "::CppType"
}
```

That just says it can only be used through a pointer that cannot be unpinned

```
type crate::MyCppWrapper {
     #cpp "::CppType"
     #layout (size = 8, align = 4)
}
```

That could mean that the type can be in the Rust or heap so long as its pinned through its lifetime.

My point is that the current description tells you how you are planning to use the type, instead of declaring how the type is restricted in one way or another.

examples/stack_owned/src/main.rs
adsnaider: I left it here for visibility. The `custom_constructor` uses the `::build()` method so I wanted to showcase both options

examples/cpp_inheritance/Cargo.toml
adsnaider: Done

book/src/call_cpp_from_rust/opaque.md
adsnaider: Done

book/src/call_cpp_from_rust/opaque.md
hkalbasi: Makes sense.

examples/stack_owned/src/main.rs
adsnaider: Done. Removed the default constructor and changed the `custom_constructor` to `new`

book/src/SUMMARY.md
```diff
@@ -16,6 +16,7 @@
   - [Writing `impl` blocks for Rust types in C++](./call_cpp_from_rust/rust_impl.md)
   - [`Box<dyn Fn>`]()
   - [Opaque C++ types](./call_cpp_from_rust/opaque.md)
+  - [Trivially Relocatable C++ types](./call_cpp_from_rust/trivially_relocatable.md)
```
hkalbasi: This will create an empty chapter I think.

zngur-generator/templates/zng_header.sptl
```diff
@@ -84,11 +84,21 @@ namespace {{ self.cpp_namespace }} {
...
   void __zngur_internal_assume_deinit(T& t) noexcept {
     __zngur_internal<T>::assume_deinit(t);
   }
 
+  template<typename T>
+  void zngur_assume_deinit(T& t) noexcept {
+    __zngur_internal<T>::assume_deinit(t);
+  }
+
   template<typename T>
```
hkalbasi: I really think these are an implementation details, and even if we need exposing more low level control, these are wrong abstraction. If C++ had private functions I would make them private. And given that examples are now independent of them, we can at least defer adding them to another PR.

HKalbasi (approved):

zngur-generator/templates/zng_header.sptl
adsnaider: Done

book/src/SUMMARY.md
adsnaider: Done

PR #120, opened 2026-04-28
Allow marking rust functions `unsafe`

adsnaider:
Note: This change requires that functions specializations are explicit on the Rust side, whereas before you could leave the specialized type out of the .zng declaration since the compiler was smart enough to figure out those specializations.

### DIFF
--- a/examples/conditional/main.zng
+++ b/examples/conditional/main.zng
@@ -46,6 +46,10 @@ mod crate {
 
 }
 
+type [crate::KeyValuePair] {
+    wellknown_traits(?Sized);
+}
+
 mod ::std {
 
     type option::Option<usize> {
@@ -76,6 +80,20 @@ mod ::std {
         fn unwrap(self) -> &crate::KeyValuePair;
     }
 
+    type option::Option<&[crate::KeyValuePair]> {
+        #layout(size = 16, align = 8);
+        wellknown_traits(Copy, Debug);
+
+        constructor None;
+        constructor Some(&[crate::KeyValuePair]);
+
+        fn unwrap(self) -> &[crate::KeyValuePair];
+    }
+
+    type ::std::ops::RangeFull {
+        #layout(size = 0, align = 1);
+        wellknown_traits(Copy);
+    }
 
     mod vec {
         type Vec<crate::KeyValuePair> {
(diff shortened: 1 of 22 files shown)


zngur-generator/src/rust.rs
```diff
@@ -950,16 +951,26 @@ pub extern "C" fn {mangled_name}("#
...
+        }
+        if is_safe {
+            let loops = (0..inputs.len()).map(|_| "loop {}").join(", ");
+            wln!(
+                self,
+                "    const {{ let _verify_safety = || {{ #[allow(unreachable_code)] let _ = {}({}); }}; }};",
+                rust_name,
+                loops
+            );
+        }
```
hkalbasi, line 971: So no check for other direction?

examples/conditional/main.zng
```diff
@@ -85,7 +85,7 @@ mod ::std {
 
             fn new() -> Vec<crate::KeyValuePair>;
             fn push(&mut self, crate::KeyValuePair);
-            fn get(&self, usize) -> ::std::option::Option<&crate::KeyValuePair> deref [crate::KeyValuePair];
+            fn get<usize>(&self, usize) -> ::std::option::Option<&crate::KeyValuePair> deref [crate::KeyValuePair];
```
hkalbasi, line 106: Now these are going to be template specialization in the C++ side instead of overloaded function? Can we still have multiple instantiation? IIRC there were some difficulties on template specialization of the methods. An example which uses multiple instantiations of a single function may help.

zngur-generator/src/rust.rs
adsnaider: No, I don't want to force that since I think there are reasons you might want to define a function as unsafe for the bridge since C++ doesn't have a borrow checker. I understand it doesn't do anything technically, but it's still good documentation

examples/conditional/main.zng
adsnaider: Done. It works fine. Also, @HKalbasi, here's another situation you need to use __zngur_internal_assume_init since as far as I can tell there's no way to construct a RangeFull in C++ (the only constructor is `RangeFull` in rust which can't be mapped to C++'s)

PR #122, opened 2026-05-29
Improve `rust::Bool` to align with rust's `bool` validity constraints

Bamboo-D:
Rust `bool` must be `0x00` or `0x01` — other byte values are invalid representations causing UB. Rust `bool` also does not support arithmetic.
This PR makes `rust::Bool` respect these constraints:

- `operator bool() const noexcept` — enable const usage, consistent with `Char::operator char32_t() const`
- `Bool() { __zngur_data[0] = 0; }` — zero-initialize to avoid invalid Rust bool representation
- `operator int() const = delete` — block implicit arithmetic via `bool->int` promotion. Deleting only `operator int()` suffices because the compiler encounters ambiguity between `operator bool` and `operator int` for all arithmetic conversions, effectively blocking `float`, `short`, etc. as well
- `friend operator<<`, `friend operator==`, `friend operator!=` — resolve overload ambiguity for `ostream` output and comparisons, caused by the deleted `operator int()` still participating in overload resolution

### DIFF
--- a/zngur-generator/templates/cpp_header.sptl
+++ b/zngur-generator/templates/cpp_header.sptl
@@ -289,11 +289,7 @@ namespace {{ self.namespace }} {
 
     public:
       {% if is_copy %}
-        {% if td.ty.path.to_string() == format!("::{}::Bool", self.namespace) %}
-        {{ name }}() { __zngur_data[0] = 0; }
-        {% else %}
         {{ name }}() { {{ alloc_heap }} }
-        {% endif %}
         ~{{ name }}() { {{ free_heap }} }
         {{ name }}(const {{ name }}& other) {
           {{ alloc_heap }}
--- a/zngur-generator/templates/cpp_header.sptl
+++ b/zngur-generator/templates/cpp_header.sptl
@@ -239,9 +239,10 @@ namespace {{ self.namespace }} {
 
     {% if td.ty.path.to_string() == format!("::{}::Bool", self.namespace) %}
       public:
-        operator bool() {
+        operator bool() const noexcept {
           return __zngur_data[0];
         }
+        operator int() = delete;
         Bool(bool b) {
           __zngur_data[0] = b;
         }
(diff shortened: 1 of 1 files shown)


hkalbasi:
Hi! Thanks for the PR. Can you please give the motivating example which was misleading or broken before this PR?

> Bool() { __zngur_data[0] = 0; }

In zngur, default constructors represent uninitialized data. We can make an exception for bool but I would prefer it to be consistent with other types, unless there is a bigger benefit for the exception.

> block implicit arithmetic via bool->int promotion

I think this is the main change in this PR, and I don't understand its motivation. What is the problem with a bool->int implicit cast? It is available for C++ bool as well.

Bamboo-D:
Thanks for the review!

- **Regarding `Bool() { __zngur_data[0] = 0; }`**
  
  The concern is that an uninitialized `rust::Bool` with a garbage byte value can cause C++ and Rust to disagree on the same value. For example:
  
  ```cpp
  rust::Bool rb;
  rb.__zngur_data[0] = 42;  // simulating garbage from uninitialized memory
  
  cpp_check_bool(rb);   // C++ sees it as true
  rust_check_bool(rb);  // Rust sees it as false
  ```
  
  This is UB in Rust, and the two sides produce contradictory results.

- **Regarding `operator int() = delete`**
  
  The idea was to make `rust::Bool` behave more like a Rust `bool`. Users can always `bool()` first if they need C++ bool semantics. But I realize this adds quite a bit of complexity. If this restriction isn't needed, I'm glad to drop it.

hkalbasi:
> This is UB in Rust, and the two sides produce contradictory results.

Yes, using uninitialized data is UB, and that contradiction is just UB in action. This is not limited to `Bool` and almost all types default constructors in zngur make objects in the uninitialized state. I don't think we should special case `Bool` here unless there is something special about it.

> The idea was to make rust::Bool behave more like a Rust bool... I'm glad to drop it.

Now I see, nice. No, keep it, I liked it. Hope it won't break too much code.

PR #123, opened 2026-06-21
Update the tutorial to the latest Zngur version

Chris00:
The tutorial did not work for me because it was not in sync with the latest Zngur version.


hkalbasi:
Thank you very much!

PR #125, opened 2026-06-25
generate enum variant accessors

GoldsteinE:
(imagine it’s a tracking issue for what was proposed in [this comment](https://github.com/hkalbasi/zngur/pull/125#issuecomment-4804586399) now)

- [x] parsing
- [x] testing
- [x] docs

variant classes:
- [x] match
- [x] match_ref
- [x] match_mut
- [x] check
- [x] constructor from fields
- [x] field access

enum class:
- [x] match
- [x] match_ref
- [x] match_mut
- [x] constructors from variants

<details><summary>original text</summary>

hi! currently, to use rust enums in cpp you have to manually write and bind functions that extract data from enum variants. this patch adds automatic generation of `.unwrap_Variant()` methods for the trivial case: tuple variants that only have a single field

this restriction avoids the question of how the return type should look for more complex cases: I think for tuple variants with multiple fields we should return `std::tuple`, and for struct variants we should codegen a helper struct, but in these cases there’re multiple options, and a single-field tuple variant case is straightforward. at the same time, this has a lot of utility, as this case is really common, and all the other cases can be emulated by putting a struct inside the single field

(unfortunately, this also generates an unwrap method for single-field tuple structs, as zng format does not distinguish between structs and enums. I could gate the generation on having multiple constructors though, let me know what you think)

obviously, this will also need tests and documentation, which I’m happy to write if you accept the idea in general

</details>

### DIFF
--- a/book/src/call_rust_from_cpp/variants.md
+++ b/book/src/call_rust_from_cpp/variants.md
@@ -1,0 +1,72 @@
+# Enum variants
+
+Enum variants can be declared with `variant name { }` syntax.
+Inside the block you can specify the fields of a variant,
+similarly to how struct fields are declared:
+
+```
+type ::std::option::Option<i32> {
+    variant None { }
+    variant Some {
+        field 0 (offset = auto, type = i32);
+    }
+}
+```
+
+Note that due to technical limitations only `auto` offsets are supported.
+
+By default, if any variants are declared, zngur will verify at compile time
+that declarations are exhaustive, i.e. no variants and no variant fields are missing. `non_exhaustive` declaration can be used to disable this check.
+
+```
+type ::std::option::Option<i32> {
+    non_exhaustive; // Some variants might be missing.
+    variant Some {
+        non_exhaustive; // Some variant fields might be missing.
+    }
+}
+```
+
+For every declared variant zngur will generate a nested class.
+If a variant is exhaustive, this class will have a constructor to build it from fields:
+
+```C++
+template<class T> using Option = rust::std::option::Option<T>;
+
+Option<int> opt = Option<int>::Some(42);
+```
+
+Each variant class has a `::match()` static method returning `std::optional` to check if the enum contains that variant:
+
+```C++
+if (auto r = Option<int>::Some::match(v)) {
+    std::cout << r->f0 << std::endl;
+}
+```
+
+`::match_ref()` and `::match_mut()` static methods can similarly be used for by-ref matching,
+returning `std::optional<Ref<_>>` and `std::optional<RefMut<_>>` respectively.
+
+These methods are only available on C++17 or higher, since they require `std::optional`.
+
+If you don't need the fields, you can also use the `::check()` static method:
+
+```C++
+if (Option<int>::None::check(v)) {
+    std::cout << "is none" << std::endl;
+}
+```
+
+Alternatively, `.match()` methods on the enum class can be used to convert the enum into `std::variant`
+of variant classes (or refs to variant classes):
+
+```C++
+std::visit(overloaded{
+    [](rust::Ref<Option<int>::None>) { std::cout << "none" << std::endl; },
+    [](rust::Ref<Option<int>::Some> s) { std::cout << "some " << s.f0 << std::endl; },
+}, v.match_ref());
+```
+
+This is also only available starting with C++17.
+
+See `examples/enums` for a runnable demonstration.
--- a/examples/char/NMakefile
+++ b/examples/char/NMakefile
@@ -1,6 +1,6 @@
 CXX = cl.exe
 # MSVC doesn't support earlier that c++14
-CXXFLAGS = /W4 /DEBUG /EHsc /std:c++14 
+CXXFLAGS = /W4 /DEBUG /EHsc /std:c++14 /Zc:__cplusplus
 WINLIBS = ntdll.lib
 
 EXAMPLE_NAME = char
--- a/examples/enums/README.md
+++ b/examples/enums/README.md
@@ -1,0 +1,10 @@
+# Example: Enums
+
+Showcases some methods of working with Rust enums.
+
+To run this example:
+
+```
+make
+./a.out
+```
--- a/examples/tutorial_cpp/main.zng
+++ b/examples/tutorial_cpp/main.zng
@@ -36,8 +36,6 @@ type crate::Item {
 
 type ::std::fmt::Result {
     #layout(size = 1, align = 1);
-
-    constructor Ok(());
 }
 
 type ::std::fmt::Formatter {
--- a/zngur-def/src/merge.rs
+++ b/zngur-def/src/merge.rs
@@ -1,7 +1,7 @@
 use crate::{
     AdditionalIncludes, ConvertPanicToException, CppRef, CppStackOwned, CppValue, LayoutPolicy,
     ZngurConstructor, ZngurExternCppFn, ZngurExternCppImpl, ZngurField, ZngurFn,
-    ZngurMethodDetails, ZngurSpec, ZngurTrait, ZngurType,
+    ZngurMethodDetails, ZngurSpec, ZngurTrait, ZngurType, ZngurVariant,
 };
 
 /// Trait for types with a partial union operation.
@@ -156,9 +156,15 @@ impl Merge for ZngurType {
         merge_by_identity(self.methods, &mut into.methods, |a, b| {
             a.data.name == b.data.name
         })?;
-        merge_by_identity(self.constructors, &mut into.constructors, |a, b| {
-            a.name == b.name
-        })?;
+        if into.constructor.is_none() {
+            into.constructor = self.constructor;
+        } else if self.constructor != into.constructor {
+            return Err(MergeFailure::Conflict(
+                "Duplicate constructor found".to_string(),
+            ));
+        }
+        into.exhaustive = self.exhaustive && into.exhaustive;
+        merge_by_identity(self.variants, &mut into.variants, |a, b| a.name == b.name)?;
         merge_by_identity(self.fields, &mut into.fields, |a, b| a.name == b.name)?;
 
         Ok(())
@@ -246,6 +252,20 @@ impl Merge for ZngurConstructor {
         if self != *into {
             return Err(MergeFailure::Conflict("Constructor mismatch".to_string()));
         }
+        Ok(())
+    }
+}
+
+impl Merge for ZngurVariant {
+    fn merge(self, into: &mut Self) -> MergeResult {
+        if self.name != into.name {
+            panic!(
+                "Attempt to merge different variants: {} and {}",
+                self.name, into.name
+            );
+        }
+        merge_by_identity(self.fields, &mut into.fields, |a, b| a.name == b.name)?;
+        into.exhaustive = self.exhaustive && into.exhaustive;
         Ok(())
     }
 }
--- a/zngur-generator/src/cpp.rs
+++ b/zngur-generator/src/cpp.rs
@@ -94,21 +94,23 @@ impl Display for CppPath {
 pub struct CppType {
     pub path: CppPath,
     pub generic_args: Vec<CppType>,
+    // TODO: is there a better way to do it?
+    pub tail: Option<String>,
 }
 
 impl CppType {
-    pub fn into_ref(self) -> CppType {
-        CppType {
-            path: CppPath::from(&*format!("{}::Ref", self.top_level_ns())),
-            generic_args: vec![self],
+    pub fn with_tail(self, tail: String) -> Self {
+        Self {
+            tail: Some(tail),
+            ..self
         }
     }
 
-    fn top_level_ns(&self) -> &String {
-        &self.path.namespace()[0]
-    }
-
     pub(crate) fn specialization_decl(&self) -> String {
+        if self.tail.is_some() {
+            panic!("specialization decl makes no sense for nested classes");
+        }
+
         let name = self.path.name();
         // Tuple<> is a little special because it is a template even though self.generic_args is empty
         if self.generic_args.is_empty() && name != "Tuple" {
@@ -120,6 +122,26 @@ impl CppType {
                 self.generic_args.iter().join(", ")
             )
         }
+    }
+
+    pub(crate) fn specialized_name(&self) -> String {
+        // We don't get `std::fmt::from_fn()` on our MSRV.
+        struct Helper<'a>(&'a CppType);
+
+        impl Display for Helper<'_> {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                write!(f, "{}", self.0.path.name())?;
+                if !self.0.generic_args.is_empty() {
+                    write!(f, "< {} >", self.0.generic_args.iter().join(", "))?;
+                }
+                if let Some(tail) = &self.0.tail {
+                    write!(f, " :: {tail}")?;
+                }
+                Ok(())
+            }
+        }
+
+        Helper(self).to_string()
     }
 
     fn header_helper(&self, state: &mut impl Write) -> std::fmt::Result {
@@ -207,12 +229,14 @@ impl From<&str> for CppType {
             None => CppType {
                 path: CppPath::from(value),
                 generic_args: vec![],
+                tail: None,
             },
             Some((path, generics)) => {
                 let generics = generics.strip_suffix('>').unwrap();
                 CppType {
                     path: CppPath::from(path),
                     generic_args: split_string(generics).map(|x| CppType::from(&*x)).collect(),
+                    tail: None,
                 }
             }
         }
@@ -347,6 +371,14 @@ impl CppMethod {
             .map(|(n, ty)| format!("{ty} i{n}"))
             .join(", ")
     }
+}
+
+#[derive(Debug)]
+pub struct CppVariant {
+    pub name: String,
+    pub constructor: Option<CppFnSig>,
+    pub match_check: String,
+    pub fields: Vec<ZngurFieldData>,
 }
 
 #[derive(Debug)]
@@ -486,7 +518,9 @@ pub struct CppTypeDefinition {
     pub ty: CppType,
     pub layout: CppLayoutPolicy,
     pub methods: Vec<CppMethod>,
-    pub constructors: Vec<CppFnSig>,
+    pub constructor: Option<CppFnSig>,
+    pub variants: Vec<CppVariant>,
+    pub discriminant: Option<String>,
     pub fields: Vec<ZngurFieldData>,
     pub from_trait: Option<RustTrait>,
     pub from_trait_ref: Option<RustTrait>,
@@ -618,7 +652,9 @@ impl Default for CppTypeDefinition {
             ty: CppType::from("fill::me::you::forgot::it"),
             layout: CppLayoutPolicy::OnlyByRef,
             methods: vec![],
-            constructors: vec![],
+            constructor: None,
+            variants: vec![],
+            discriminant: None,
             fields: vec![],
             wellknown_traits: vec![],
             from_trait: None,
--- a/zngur-generator/src/rust.rs
+++ b/zngur-generator/src/rust.rs
@@ -29,6 +29,7 @@ impl IntoCpp for RustPathAndGenerics {
                 .chain(named_generics)
                 .map(|x| x.into_cpp(namespace, crate_name))
                 .collect(),
+            tail: None,
         }
     }
 }
@@ -48,6 +49,7 @@ impl IntoCpp for RustTrait {
                     .chain(Some(&**output))
                     .map(|x| x.into_cpp(namespace, crate_name))
                     .collect(),
+                tail: None,
             },
         }
     }
@@ -96,6 +98,7 @@ impl IntoCpp for RustType {
             RustType::Boxed(t) => CppType {
                 path: CppPath::from(&*format!("{namespace}::Box")),
                 generic_args: vec![t.into_cpp(namespace, crate_name)],
+                tail: None,
             },
             RustType::Ref(m, t) => CppType {
                 path: match m {
@@ -103,10 +106,12 @@ impl IntoCpp for RustType {
                     Mutability::Not => CppPath::from(&*format!("{}::Ref", namespace)),
                 },
                 generic_args: vec![t.into_cpp(namespace, crate_name)],
+                tail: None,
             },
             RustType::Slice(s) => CppType {
                 path: CppPath::from(&*format!("{namespace}::Slice")),
                 generic_args: vec![s.into_cpp(namespace, crate_name)],
+                tail: None,
             },
             RustType::Raw(m, t) => CppType {
                 path: match m {
@@ -114,6 +119,7 @@ impl IntoCpp for RustType {
                     Mutability::Not => CppPath::from(&*format!("{namespace}::Raw")),
                 },
                 generic_args: vec![t.into_cpp(namespace, crate_name)],
+                tail: None,
             },
             RustType::Adt(pg) => pg.into_cpp(namespace, crate_name),
             RustType::Tuple(v) => {
@@ -126,6 +132,7 @@ impl IntoCpp for RustType {
                         .into_iter()
                         .map(|x| x.into_cpp(namespace, crate_name))
                         .collect(),
+                    tail: None,
                 }
             }
             RustType::Dyn(tr, marker_bounds) => {
@@ -140,6 +147,7 @@ impl IntoCpp for RustType {
                                 .map(|x| CppType::from(&*format!("{namespace}::{x}"))),
                         )
                         .collect(),
+                    tail: None,
                 }
             }
             RustType::Impl(_, _) => panic!("impl Trait is invalid in C++"),
@@ -423,11 +431,6 @@ fn mangle_name(name: &str, mangling_base: &str) -> String {
         w!(name, "{}{pos}", which.2);
     }
     name
-}
-
-pub struct ConstructorMangledNames {
-    pub constructor: String,
-    pub match_check: String,
 }
 
 impl RustFile {
@@ -732,13 +735,12 @@ pub extern "C" fn {constructor}("#
         constructor
     }
 
-    pub fn add_constructor(
+    pub fn add_constructor<'a>(
         &mut self,
         rust_name: &str,
-        args: &[(String, RustType)],
-    ) -> ConstructorMangledNames {
+        args: impl IntoIterator<Item = (&'a String, &'a RustType)> + Clone,
+    ) -> String {
         let constructor = self.mangle_name(rust_name);
-        let match_check = format!("{constructor}_check");
         w!(
             self,
             r#"
@@ -746,7 +748,7 @@ pub extern "C" fn {constructor}("#
 #[unsafe(no_mangle)]
 pub extern "C" fn {constructor}("#
         );
-        for (name, _) in args {
+        for (name, _) in args.clone() {
             w!(self, "f_{name}: *mut u8, ");
         }
         w!(
@@ -758,6 +760,11 @@ pub extern "C" fn {constructor}("#
             w!(self, "{name}: ::std::ptr::read(f_{name} as *mut {ty}), ");
         }
         wln!(self, "}}) }} }}");
+        constructor
+    }
+
+    pub(crate) fn add_match_check(&mut self, rust_name: &str) -> String {
+        let match_check = self.mangle_name(&format!("{rust_name}_check"));
         w!(
             self,
             r#"
@@ -767,10 +774,44 @@ pub extern "C" fn {match_check}(i: *mut u8, o: *mut u8) {{ unsafe {{
     *o = matches!(&*(i as *mut &_), {rust_name} {{ .. }}) as u8;
 }} }}"#
         );
-        ConstructorMangledNames {
-            constructor,
-            match_check,
+        match_check
+    }
+
+    pub(crate) fn add_discriminant(
+        &mut self,
+        rust_name: &str,
+        variants: &[ZngurVariant],
+    ) -> Option<String> {
+        if variants.is_empty() {
+            return None;
         }
+
+        let name = self.mangle_name(&format!("{rust_name}::match"));
+        w!(
+            self,
+            r#"
+#[allow(non_snake_case)]
+#[unsafe(no_mangle)]
+pub extern "C" fn {name}(i: *mut u8) -> u32 {{ unsafe {{
+    match &*(i as *mut {rust_name} as *const _) {{"#
+        );
+        for (n, variant) in variants.iter().enumerate() {
+            let name = &variant.name;
+            w!(
+                self,
+                r#"
+        {rust_name}::{name} {{ .. }} => {n},
+        "#
+            );
+        }
+        w!(
+            self,
+            r#"
+    }}
+}} }}
+            "#
+        );
+        Some(name)
     }
 
     pub(crate) fn add_field_assertions(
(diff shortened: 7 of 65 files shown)


GoldsteinE:
out of interest, I tried to also allow multi-field tuple variants, but doing that via `std::tuple` requires modifications to cpp function template: there’s no good way to create an `std::tuple` on rust side, so cpp wrapper needs to pass separate pointer for every field. I’ll think about it some more, but I’m not yet sure what the best way to achieve this.

hkalbasi:
Hi! Thanks for the PR. I think you are addressing a real problem, I thought about this problem a little, let's explore the problem space.

The problem is that there is Rust's `match` (and it's syntax sugars `if let`, `let else`, ...) which C++/Zngur doesn't have, so using of ADT types is limited to Rust defined methods.

The `match` syntax does these things:
1. Switch casing over variants
2. Checking if a variant is match, i.e. `matches!(value, Variant { .. })`
3. Moving out of fields of variants
4. Getting `&`/`&mut` reference of variant fields from `&`/`&mut value`

If I want to design a solution addressing all of these on the fly, I would add a class member for each variant into the enum, and each variant class has these static functions on it:
```
1. match(Enum) -> Variant
2. match_ref(Ref<Enum>) -> Ref<Variant>
3. match_mut(RefMut<Enum>) -> RefMut<Variant>
4. check(Ref<Enum>) -> bool
5. Constructor from fields
```
And enum will get these static function members:
```
1. match(Enum) -> std::variant<Variant1, Variant2, Variant3>
2. Constructor from all variants
```
These variant classes can get field access with similar trick that we use for structs. Inside them, there is only an `Enum` data member which is the `Enum` that we know is in the variant state. So your `unwrap_Variant` would become something like this:
```
Field f = e.unwrap_Variant();
// Becomes
Field f = Enum::Variant::match(e).f0;
```
What do you think about this? I generated this on the fly, so probably there is something I missed. And names are definitely subject of bikeshed.

I would like to determine the general solution first, then if you want to implement only a part of it (e.g. for single field variants) that's fine as long as it's compatible with the big picture plan.

GoldsteinE:
hi! thanks for you thoughts.

do I understand correctly that `Enum::Variant` would hold the `Enum`-sized byte array, just like the `Enum` itself does? (and then `match_ref` and `match_mut` are just check + pointer cast). I think I can try implementing this, although I’ll need to research how you do fields (I hope I can reuse some of the code generating infrastructure, since generating a variant class should be really similar to generating a struct).

this syntax is nice and general, although it might look a bit more verbose, e.g.

```cpp
if (val.matches_Variant()) {
    auto val = val.unwrap_Variant();
}
```

turns into

```cpp
if (Enum::Variant::check(val)) {
    auto val = enum::Variant::match(val).f0;
}
```

but I don’t think it’s a big deal. either way, I’m happy to try implementing (some subset of) your design (probably starting with `check()` and `match()`, since that’s what I need right now lol, but I’ll try to get more methods done)

GoldsteinE:
also would probably need `.match_ref()` and `.match_mut()` on enum, corresponding to `match &en` and `match &mut en`?

GoldsteinE:
I want to also share my thoughts on implementation so you can verify if that’s reasonable.

1. I want to add a new variant to `RustType` representing individual enum variants. based on that I’ll generate a faux `ZngurType` to reuse the field & constructor generating infrastructure. this faux type inherits layout from the parent. this covers field access and constructor from fields.
2. this does not cover checking and matching methods, which need to be generated separately. all the `match{_ref,_mut,}` methods on variant types seem relatively straightforward and basically compile into check + transmute. I’ll add new methods to `RustFile` for that, and corresponding code to `cpp_header.sptl`. (if I undestand correctly, basically all the C++ code generation happens via askama, as opposed to Rust code generation, which mostly uses `write!()`)
3. all the static functions on enum need to be written manually, again in `cpp_header.sptl`. I think no new code needs to go into C++ source file.

does that look right? if so, I’ll start implementing that.

hkalbasi:
> do I understand correctly that Enum::Variant would hold the Enum-sized byte array

It can hold the `Enum` itself, saving some code generation about move constructor and similar things.

> this syntax is nice and general, although it might look a bit more verbose

What about also providing `match_opt`, `match_opt_ref` and `match_opt_mut` returning `std::optional`? Then your example becomes:
```
if (auto val = Enum::Variant::match_opt(val)) {
    auto val = val->f0;
}
```

> also would probably need .match_ref() and .match_mut() on enum, corresponding to match &en and match &mut en?

Makes sense.

> does that look right? if so, I’ll start implementing that.

I think you need to change the zng file as well, for annotating fields. A candidate syntax:
```
type Option<i32> {
    variant Some {
        constructor(i32);
        field 0 (offset = 4, type = i32);
    }
    variant None {
         constructor;
    }
}
```
Other than that, I think it is good for start.

GoldsteinE:
> What about also providing match_opt, match_opt_ref and match_opt_mut returning std::optional?

maybe, yeah, although the `->f0` part of verbosity is still there. I think it’s also possible to generate `std::get_if` overloads if we want it to feel more C++-ish.

> I think you need to change the zng file as well, for annotating fields. A candidate syntax:

it is a bit unfortunate that this would require mentioning every field twice. maybe variants could have implicit constructors based on their fields?

GoldsteinE:
one other interesting moment: if not every variant is declared (do we want to support this?), `Enum::match()` needs *something* to do in the default case. should it just panic? should this be a compile error? maybe we could have a separate `Default` pseudo-variant as a catch-all.

hkalbasi:
> it is a bit unfortunate that this would require mentioning every field twice. maybe variants could have implicit constructors based on their fields?

This also exists for struct. We could have a `constructor auto` or something like that to tell the Zngur generate a constructor based on available fields. But anything we do should be consistent with structs. The idea for structs was that you may want the constructor, but couldn't afford offsets for some fields.

> one other interesting moment

I think we have these options:
1. Give up on `match` on enum, at least for now
2. Runtime panic/exception when this happens
3. Force all variants to be available by a compile error
4. Have a `Default` variant unconditionally
5. Add some annotation, controlling whether `match` functions need to be generated, and if so, controlling whether it is exhaustive and generate `Default` variant only if it is needed

I don't like 2 and 4, and going with 1 makes 3 in future a breaking change (breaking change isn't a hard limit, but I prefer to avoid it).

GoldsteinE:
the difference with structs is that enum variants are guaranteed to have all fields public. could you explain what do you mean by “couldn’t afford offsets”? I thought offsets only incur runtime cost if they’re used, and otherwise are just a single `usize` somewhere in the binary.

going with 3 for now is the most conservative choice probably? also while 1 makes 3 a breaking change, it still leaves 5 open (although that’s more boilerplate forever).

hkalbasi:
> I thought offsets only incur runtime cost if they’re used, and otherwise are just a single usize somewhere in the binary.

Offsets were mandatory at the time, we then got runtime offsets later. So we may want to revisit this (but I still want consistency between structs and variants).

> the difference with structs is that enum variants are guaranteed to have all fields public.

If a field is not public for a struct, the constructor is not usable.

> going with 3 for now is the most conservative choice probably? also while 1 makes 3 a breaking change, it still leaves 5 open (although that’s more boilerplate forever).

:+1: . It is not forever (yet), it just creates some barrier for adopting 1. I'm not afraid of doing breaking changes (this PR is going to be a breaking change, breaking `constructor Some(i32)`) I just want to not do it for no reason.

GoldsteinE:
> If a field is not public for a struct, the constructor is not usable.

yeah, and that is not a concern for enum variants, because all the fields are always public, so you can always specify every field. on the other hand, you may not be able to specify some struct field, in which case we can’t auto-generate a constructor.

hkalbasi:
> yeah, and that is not a concern for enum variants, because all the fields are always public, so you can always specify every field. on the other hand, you may not be able to specify some struct field, in which case we can’t auto-generate a constructor.

This is fine for an explicit `constructor auto` since user will just get a compile error and remove that line, but if we want to implicitly generate constructors for all enum variants, then we need to diverge. If you want to go that path and only have `field` annotations for variants, then it's fine to not making it consistent with structs.

GoldsteinE:
I think implicitly generating constructors for enum variants makes sense, but if you want full consistency, I can try implementing `constructor auto`

hkalbasi:
I'm fine with implicit constructor for variants, and not touching structs in this PR.

GoldsteinE:
status update: making `Variant` a `ZngurType` turned out to be kind of annoying, since `ZngurType` has a lot of stuff that’s totally irrelevant for variants, including `layout`. maybe it would be easier to diverge somewhere down the line, and make variants a separate struct.

side note: is there any way to improve `cargo check` times on `zngur-parser`? it takes ~3.5 minutes on my laptop (apparently spent in borrowck), and it turns out I was quite used to interactive on-the-fly diagnostics and fast `cargo test` loop lol

GoldsteinE:
> this PR is going to be a breaking change, breaking constructor Some(i32))

do I understand correctly that I should remove enum constructors completely, replacing them with variant declarations (rather than letting the two coexist)?

GoldsteinE:
status update update: made a checklist tracker in the original post.

I need to do field support now, and I’m a bit intimidated by all the machinery. I definitely don’t want to duplicate all this code, so ideally it would be generic for type fields and enum variant fields. I think basically the only thing that would change between type fields and variant fields is the base pointer, right? could you point me to where this base pointer is created/passed?

hkalbasi:
>  is there any way to improve cargo check times on zngur-parser? it takes ~3.5 minutes on my laptop (apparently spent in borrowck), and it turns out I was quite used to interactive on-the-fly diagnostics and fast cargo test loop lol

~You can download a faster laptop with more ram.~ The problem is the chumsky crate we use for parser. The [chumksy has some advice on this topic](https://docs.rs/chumsky/latest/chumsky/guide/_00_getting_started/index.html#compilation-times) that you can try.

> do I understand correctly that I should remove enum constructors completely, replacing them with variant declarations (rather than letting the two coexist)?

I think so.

> could you point me to where this base pointer is created/passed?

The black magic trick is that you create an annonymous union of the data and all fields. So the field would get the same address of the struct/variant data. So you can cast `this` to `Variant*`/`Enum*` and expect it to work.

GoldsteinE:
okay, thanks, I’ll try adapting all this magic so it hopefully also works with variant classes without separate implementation

> The black magic trick is that you create an annonymous union of the data and all fields. So the field would get the same address of the struct/variant data.

this is cool, but isn’t it kinda union punning between field types, which should be UB in C++?.. or is the fact that no actual memory access happens saves us here?

hkalbasi:
> is the fact that no actual memory access happens saves us here

I think so. The act of taking pointer to inactive members is not UB, but dereferencing that pointer is UB (which we don't).

GoldsteinE:
okay, so the approach used for structs unfortunately won’t work for enums, because `offset_of!()`, required both for assertions with static offsets and for auto offsets [does not work on enums](https://github.com/rust-lang/rust/issues/120141). barring absolutely ridiculous shenanigans like [abusing the optimizer and then extracting this info by passing an uninit value](https://godbolt.org/z/7Y5dG8Gaf) I think we straight up can’t get access to these offsets without having an enum value.

there’re two worse versions that I see, but maybe there’s something better I’m missing:

1. we could calculate them lazily on first access. this makes first field access unpredictably longer and turns a regular load into atomic relaxed load + (very predictable) branch for subsequent accesses. dirty, but probably not *that* awful?

2. we could instead generate projection functions mapping a reference to the variant to a pointer to the field. I’m not sure if it’s more or less overhead, since it’ll be an FFI call for every field every time, but no atomics and no conditionals. also all the accesses take the same time in this case.

in either case, we could probably reuse most of the logic still, as basically only offset calculation changes. also, in either case there’s seemingly no point in supporting non-auto offsets, as we have no real way of checking them in compile time (well, I guess we could if the enum implements `Default` or something, but this feels really weird).

of these two options, I think option 2 is more sensible, as it doesn’t introduce mutable state.

GoldsteinE:
I pushed a version that covers all the API surface you described. the tests will fail rn, since I yeeted named constructors in favour of variants, but otherwise it should be ready. this is a semi-large patch, so I pushed the implementation while I work on tests and docs.

GoldsteinE:
status update: porting `rustyline` example kinda forces a decision on exhaustiveness. fully describing its `Error` type is way too messy.

interestingly, there’re two kinds of nonexhaustiveness:
- zng could skip describing some *variants*, in which case we can’t generate `Enum::match()`, but can generate basically everything else
- or it could skip describing some variant *fields*, in which case we can generate `Enum::match()`, but can’t auto-generate a constructor

I propose a new `non_exhaustive` annotation (mirroring rust’s attribute) that covers both cases: when applied on type-level, it disables `Enum::match()` generation, and when applied on variant-level it disables variant constructor generation

(alternatively, it could add a `Default` option into variant, as you proposed)

hkalbasi:
I like omitting the `match` function more than `Default`. People can always use `if Var1::match else if Var2::match`.

GoldsteinE:
status update: implemented `non_exhaustive`, fixed every (hopefully) example, fixed a bunch of bugs. would be nice to have a CI run now, because I didn’t actually test `tutorial-wasm32` (I have a very-much non-`rustup` setup).

I will do book changes next, although I’m not sure it would be very soon, since I’ll be away for some time.

GoldsteinE:
another note: migrating examples revealed some usability papercuts. I think the most annoying by far is that when you do stuff like

```cpp
auto meow = Enum::Variant(42);
```

`meow` has type `Enum::Variant`, not type `Enum`, and so has none of its methods. you need to write out

```cpp
Enum meow = Enum::Variant(42);
```

which is annoyingly repetitive (and it’s not intuitive that this is what you need to do from the error). I’m not sure what to do with this. one band-aid would be to provide some kind of upcast method, like `Enum::Variant(42).bikeshed()` or smth, but that’s still not pretty. the same problem goes for lambdas, where returning different variants from different branches leads to inference failures.

some other issues:
- `x.matches_Variant()` is much shorter than `Enum::Variant::check(x)`, we might want to restore that? same with `Enum::Variant::match(x)`, a method notation might be desirable.
- in template functions `Enum::Variant` is a dependent name, requiring `typename Enum::Variant` (but at least the compiler tells you what to do)

GoldsteinE:
as a wild idea, variants might inherit from the enum class instead of containing it, but that would need a different solution for fields, since we can’t put a superclass into a union. might have other problems, C++ inheritance is not something I’m very familiar with

GoldsteinE:
thanks! I somehow missed a test I guess, I’ll re-check

hkalbasi:
> migrating examples revealed some usability papercuts

I thought about this problem. The inheritance is dead end (It breaks the union trick, won't fix the lambda problem, make upcasting ambiguous, adds a vtable member to every enum, ...) but I had thought of two more solutions you didn't mention:
* Repeat all methods for every variant as well. Still won't fix the lambda problem, and makes a n*m code generation problem (which is bad due compile times).
* Return `Enum` directly. So we could not use the constructor and need a `build` static function.

I concluded that `Enum(Enum::Variant(1))` was not that bad, so I didn't bring it up. I still think `Enum::Variant(1).bikeshed()` is not much better than `Enum(Enum::Variant(1))` and given that the latter already works, do we need the bikeshed method? I don't like my solutions either, and would say the status quo is good enough, but don't have a strong opinion here so brought them up so that .

> x.matches_Variant() is much shorter than Enum::Variant::check(x), we might want to restore that? same with Enum::Variant::match(x), a method notation might be desirable.

I prefer to not have methods on types, to prevent shadowing original methods. `Enum::Variant::check(x)` only has the additional `Enum::` which could be `using` ed away if it becomes repeated.

> in template functions Enum::Variant is a dependent name, requiring typename Enum::Variant (but at least the compiler tells you what to do)

I hate C++. Probably we can't do much here.

> I will do book changes next, although I’m not sure it would be very soon, since I’ll be away for some time.

That's fine, and thanks for the great work on this PR!

GoldsteinE:
> I still think Enum::Variant(1).bikeshed() is not much better than Enum(Enum::Variant(1)) and given that the latter already works, do we need the bikeshed method?

I think my example here hid the problem. consider:

```cpp
Result<Unit, ReadlineError>(Result<Unit, ReadlineError>::Err(err))
```

this is really annoying to write, doubly so because irl it would be in some kind of namespace, requiring either a lot of `using` or even more typing.

the same goes for all the other free functions.

> to prevent shadowing original methods

I understand this concern, but I feel like a weird enough naming convention (like the original `check_Variant`, which mixed snake and camel case) may be enough to avoid this problem in practice.

GoldsteinE:
noooo, not MSVC т_т

GoldsteinE:
that is really fun! I love microsoft visual studio c++ compiler!

https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/

hold on a sec, I’ll fix all the nmakefiles

GoldsteinE:
oh, right, this was a thing too. sorry to make you re-approve it so many times, I just don’t have MSVC locally

hkalbasi:
> oh, right, this was a thing too. sorry to make you re-approve it so many times, I just don’t have MSVC locally

No worries. It's such a dumb feature of github.

zngur-generator/src/rust.rs
```diff
@@ -764,10 +771,46 @@ pub extern "C" fn {match_check}(i: *mut u8, o: *mut u8) {{ unsafe {{
...
+        rust_name: &str,
+        variants: &[ZngurVariant],
+    ) -> Option<String> {
+        if variants.is_empty() {
+            return None;
+        }
+
+        let name = self.mangle_name(&format!("{rust_name}::match"));
+        // REVIEW: this straight up returns u32 instead of doing the output pointer dance
+        // is there some reason not to do it like this?
```
hkalbasi: It's fine when we know the output is u32. Also, we can use a simple pointer instead of a pointer to a pointer. The out pointer and pointer to pointer is for when we want to be generic over everything.

examples/char/NMakefile
```diff
@@ -1,6 +1,6 @@
 CXX = cl.exe
 # MSVC doesn't support earlier that c++14
-CXXFLAGS = /W4 /DEBUG /EHsc /std:c++14
```
hkalbasi, line 3: Let's just add a `|| !defined(__cplusplus)` to our guards, to not requiring a new flag for windows people, and supporting other broken compilers.

examples/enums/README.md
```diff
@@ -0,0 +1,10 @@
+# Example: Simple
```
hkalbasi: The name is copy paste.

examples/tutorial_cpp/main.zng
```diff
@@ -36,8 +36,11 @@ type crate::Item {
 
 type ::std::fmt::Result {
     #layout(size = 1, align = 1);
+    non_exhaustive;
 
-    constructor Ok(());
+    variant Ok {
```
hkalbasi: Is this even used?

zngur-generator/src/cpp.rs
```diff
@@ -94,13 +94,27 @@ impl Display for CppPath {
...
     pub path: CppPath,
     pub generic_args: Vec<CppType>,
+    // TODO: is there a better way to do it?
+    pub tail: Option<String>,
 }
 
 impl CppType {
     pub fn into_ref(self) -> CppType {
+        if self.tail.is_some() {
+            // TODO: this will do `Option::Some` -> `Ref<Option::Some>` conversion in the future
```
hkalbasi: So what happens when we call match_ref currently? Does zngur panics?

zngur-generator/src/rust.rs
```diff
@@ -797,6 +840,35 @@ pub static {mn}: usize = ::std::mem::offset_of!({owner}, {name});
...
 
+    pub(crate) fn add_variant_field_calculations(
+        &mut self,
+        field: &ZngurField,
+        owner: &RustType,
+        variant: &str,
+    ) -> String {
+        let ZngurField { name, .. } = field;
+        let mn = self.mangle_name(&format!("{owner}_{variant}_field_{name}_offset"));
+        // REVIEW: what should we do if not matched? this should not happen,
```
hkalbasi: We control this and nothing bad could call it, right? If so, do a `unsafe { unreachable_unchecked() }` which helps the optimizer optimize this function to a single constant.

zngur-generator/templates/cpp_header.sptl
```diff
@@ -394,14 +424,71 @@ namespace {{ self.namespace }} {
...
-    {% for constructor in td.constructors %}
+    {% if let Some(constructor) = td.constructor %}
       {{ td.ty.path.0.last().unwrap() }}(
         {{ splat!(&constructor.inputs, |n, ty|, "{ty}") }}
       ) noexcept ;
+    {% endif %}
+
+#if __cplusplus >= 201703L
+    {% if let Some(discriminant) = td.discriminant %}
+      // REVIEW: you wrote it should be a static method, but I took the liberty of making it a normal method instead.
```
hkalbasi: Yes I meant a normal method. `match_ref` and `match_mut` are not keywords, but they aren't there?

hkalbasi: I'm fine with those being a normal method as well. My requirements are a bit fuzzy :). Also, using static methods won't help in avoiding name conflict since ~C++ is garbage~ static methods are also available as `obj.method` syntax. Function overloading helps us here, but relying on function overloading for preventing conflicts makes things confusing.

zngur-generator/templates/cpp_header.sptl
```diff
@@ -594,6 +722,17 @@ namespace {{ self.namespace }} {
       {% endif %}
     {% endfor %}
 
+#if __cplusplus >= 201703L
+    {% if let Some(discriminant) = td.discriminant %}
+      // REVIEW: it could be also called `match` in principle?
```
hkalbasi: You mean calling both functions `match` by function overloading? Then we would lose `Enum e; e.match_mut()` but the conflict prevention is nice.

examples/char/NMakefile
GoldsteinE: it’s defined, it’s just set to 1997. we could check for `_MSVC_LANG`, which should work regardless of flags, although it’s really annoying that we have to do this.

zngur-generator/src/cpp.rs
GoldsteinE: nah, this is just unreachable I think, `.match_ref()` doesn’t use this func. (but maybe it should, I’ll check if that’s possible)

zngur-generator/src/rust.rs
GoldsteinE: yeah, I’m just reluctant to introduce UB like that, so I did an abort for now. I could do unreachable_unchecked if you prefer that, we do control this and this should be statically prevented from happening.

zngur-generator/templates/cpp_header.sptl
GoldsteinE: it’s not really function overloading I guess, because the methods are defined on distinct types, but yeah, that does lose this, and it might be to big of an ergonomic hit

examples/char/NMakefile
hkalbasi: It doesn't make any sense. So windows people deserve this misery I guess and we shouldn't help them then.

zngur-generator/src/rust.rs
hkalbasi: We will get the UB anyway when we move to `offsetof!`, so I think the risk of UB here is justified.

zngur-generator/templates/cpp_header.sptl
hkalbasi: So let's keep them as is (I think `Enum e; e.match_mut()` already doesn't work in this PR, please add it and use it in tests).

examples/enums/README.md
GoldsteinE: Fixed this README.

zngur-generator/src/rust.rs
GoldsteinE: Removed an extra pointer level, cleaned up the REVIEW comment.

examples/tutorial_cpp/main.zng
GoldsteinE: No, it’s not, removed the variants and `non_exhaustive`.

zngur-generator/src/rust.rs
GoldsteinE: Changed it to `unreachable_unchecked()`, left a SAFETY comment.

examples/char/NMakefile
GoldsteinE: just to be clear, what should I do in this PR? should I leave the extra flag in NMakefiles?

GoldsteinE:
okay, I’m back(-ish). I addressed the nits and will clean up the match methods now (after rebasing, since github shows conflicts)

GoldsteinE:
oh, I think the conflict with template types is semantic, that’s unfortunate. I’ll have to research this code to make enums compatible.

zngur-generator/src/cpp.rs
GoldsteinE: not really possible with the `tail` hack and I don’t know a better solution for that. this is very much unreachable though, since `CppType::into_ref()` is dead code. I yeeted the function and `xtask ci` still passes.

zngur-generator/templates/cpp_header.sptl
```diff
@@ -846,8 +1029,131 @@ struct __zngur_internal< {{ ref_kind }} < {{ td.ty }} > > {
         ::{{ self.namespace }}::__zngur_internal_assume_deinit(i{{ n }});
       {% endfor %}
     }
+  {% endif %}
+
+  {% for variant in td.variants %}
+    // REVIEW: missing space after noexcept ensures it's unconditionally noexcept,
```
GoldsteinE: thread. I think I carefully managed spaces around `noexcept`s, but this is a kind of non-intuitive way to do that.

zngur-generator/templates/cpp_header.sptl
GoldsteinE: added & tested `.match_{ref,mut}()` for enum classes

GoldsteinE:
I rebased onto the latest main & addressed all the review comments (would be nice to get a CI run at this point in case something broke in MSVC or smth).

the one remaining action point is docs. I think my plan is to add a “Variants” chapter right after “Fields”, and then just mechanically replace every other relevant instance of `constructor` with `variant`, since I don’t think book actually explains enum constructors anywhere currently. please ack if that sounds sensible to you and I’ll do it today or tomorrow.

<details><summary><code>rg constructor book</code></summary>

```
book/src/template_types.md
11:    constructor Some(T);
12:    constructor None;
20:    // the constructors and methods are automatically defined

book/src/how_it_compares.md
93:calling their move and copy constructors using `=` operator,

book/src/zngur.md
111:  // `make_box` is similar to the `make_unique`, it takes constructor arguments

book/src/tutorial.md
214:To fix that, we need to add the constructor for the `Item`:
224:    constructor { name: ::std::string::String, size: u32 };
228:But it doesn't solve the problem, since we can't create a `String` so we can't call the constructor.
564:    constructor Ok(());

book/src/philosophy.md
50:### Copy constructors are deleted, manual `.clone()` should be used
52:Implicit copy constructors are a source of accidental performance cost,
57:Zngur doesn't delete the copy constructor.
61:Normally in C++ the default constructor creates a basic initialized object,
70:But in Zngur, default constructor always exists and creates an uninitialized object, so this code is invalid:
77:An alternative would be running `Default::default()` in the default constructor.
91:If `Vec<int32_t> v` used the default constructor,

book/src/import.md
76:        constructor None;
77:        constructor Some(i32);
163:        constructor None;
164:        constructor Some(i32);

book/src/how_it_works.md
152:since it might need a nontrivial move constructor incompatible with Rust moves.

book/src/call_rust_from_cpp/name_mapping.md
50:That is, the only valid way to create a `bool` is using it's standard constructors.
91:        constructor V4(Ipv4Addr);
92:        constructor V6(Ipv6Addr);

book/src/call_rust_from_cpp/layout_policy.md
66:This policy deletes the constructor of the type,

book/src/call_rust_from_cpp/wellknown_traits.md
9:Adding this trait enables the copy constructor for the type on the C++ side,

book/src/call_cpp_from_rust/opaque.md
165:instance, a a move constructor which can use the fact that the provided
167:constructor can efficiently steal the underlying data.
177:Furthermore, C++ allows you to `= delete` a move constructor which effectively

book/src/call_rust_from_cpp/special_types.md
7:This type also has a constructor from C++ `bool`
```

</details>

examples/char/NMakefile
hkalbasi: Yes I think the current state is ok.

GoldsteinE:
I wrote the book chapter (and fixed other examples). with that, I think, this should be fully ready for review. I cleaned up the comment history, so it is in a mergeable state.

book/src/call_rust_from_cpp/variants.md
```diff
@@ -0,0 +1,66 @@
...
+For every declared variant zngur will generate a nested class.
+If a variant is exhaustive, this class will have a constructor to build it from fields:
+
+```C++
+template<class T> using Option = rust::std::option::Option<T>;
+
+Option<int> opt = Option<int>::Some(42);
+```
+
+Each variant class has a `::match()` static method to check if the enum contains that variant:
```
hkalbasi: Please mention that these methods return what, e.g. `std::optional<Ref<Variant>>`.

zngur-def/src/merge.rs
```diff
@@ -250,6 +255,19 @@ impl Merge for ZngurConstructor {
...
 
+impl Merge for ZngurVariant {
+    fn merge(self, into: &mut Self) -> MergeResult {
+        if self.name != into.name {
+            panic!(
+                "Attempt to merge different variants: {} and {}",
+                self.name, into.name
+            );
+        }
+        merge_by_identity(self.fields, &mut into.fields, |a, b| a.name == b.name)?;
```
hkalbasi, line 267: I guess we need to handle `exhaustive` field here (probably by `&&` ing them?) as well + a test about merging variants.

book/src/call_rust_from_cpp/variants.md
hkalbasi: And that due to `std::optional` these methods are gated in C++17.

hkalbasi (approved):

zngur-generator/templates/cpp_header.sptl
hkalbasi: I don't see the problem. Can you please explain more?

GoldsteinE: `"noexcept "` (with trailing space) gets removed in postprocessing if panic-to-exception is enabled, `noexcept` without trailing space doesn’t. some methods are unconditionally unexcept, while some are only noexcept when panics don’t throw exceptions. managing the subtle difference betweeen `noexcept{` and `noexcept {` is a bit unintuitive.

hkalbasi: Yeah, it is really fragile, but I guess it is fine for now.

zngur-def/src/merge.rs
GoldsteinE: > a test about merging variants

do you mean adding it into `simple_import` example? I grepped for `#[test]`, and couldn’t find merge-specific tests. or do you mean adding it to parser tests?

hkalbasi: Yes I meant an example. I'm not sure if the simple_import example actually uses the merge logic, but there is an example for it (also the new template thing uses merge logic, you can add there as well)

We eventually need some end to end tests, but we currently (ab)use examples.

book/src/call_rust_from_cpp/variants.md
GoldsteinE: done

zngur-def/src/merge.rs
GoldsteinE: done, https://github.com/hkalbasi/zngur/compare/f4c1b1f1d95abec41cbc50ec08c6d2cd69107b33..03d6a106868b4886202d96e985059c8e4613262d.

for the comprehensive merging test I added a second file into `enums`, but I also tinkered a bit with the `template_types` test so the template-caused merging is also tested. (I discovered and reported an unrelated bug in the process)

zngur-generator/templates/cpp_header.sptl
GoldsteinE: I removed the REVIEW comment then.

hkalbasi:
Thank you very much!

PR #126, opened 2026-07-08
allow renaming of methods on the c++ side

pennae:
this lets us expose methods overloaded on their return type only (such as `AsRef::as_ref`) without wrapper traits on the rust side for merely renaming and reëexposing of a method. the c++ name is encoded into the rust wrapper function even when no renaming is needed for consistency; we could skip this step since the mangling will be uniqure either way, but always producing the same format of mangled name seems much nicer.

### DIFF
--- a/book/src/call_rust_from_cpp/name_mapping.md
+++ b/book/src/call_rust_from_cpp/name_mapping.md
@@ -67,3 +67,33 @@ char* bc = (char*)&b;
 So Zngur can't use `bool` the way it uses `uint32_t` and friends.
 But it adds some special codes to `bool`
 so that you can use it in an `if` statement or cast it to and from C++ `bool`.
+
+## Renaming generated C++ wrapper functions
+
+Traits may be implemented multiple times with different parameters for the same Rust type, providing identically named methods with different type signatures.
+C++ does not support overloading functions based on their return type in this manner, but we still want to be able to expose all implementations of a trait without requiring *additional* traits just to rename methods.
+Zngur allows renaming an exposed method on the C++ side from the spec itself to support this.
+Since C++ does support overloading methods based on their argument types such renaming is only needed when method implementations differ in their return types only.
+
+```
+type ::std::string::String {
+    #layout(size = 24, align = 8);
+
+    fn as_ref(&self) -> &str use ::std::convert::AsRef;
+    // renaming needed for the second overload since only the return type differs
+    fn as_ref(&self) -> &[u8] use ::std::convert::AsRef as as_ref_u8;
+}
+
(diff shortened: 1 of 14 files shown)


book/src/call_rust_from_cpp/overloaded_traits.md
```diff
@@ -0,0 +1,29 @@
+# Overloaded traits
```
hkalbasi: I think it would be better for this to be a subsection of another file, e.g. name mapping.

hkalbasi (approved):
Thanks for the PR!

book/src/call_rust_from_cpp/overloaded_traits.md
pennae: that's fair :) we've moved the summary line to still keep the feature immediately discoverable from the overview though rather than merging it into the name mapping file itself though; that section starting with *type* name mappings feels like it'd make readers skip over the page a bit too quickly if they're looking for method names instead. we can merge it into the name mapping file entirely too though, no strong feelings either way

hkalbasi: I think for discoverability we can detect when we are going to emit invalid C++ code, and emit a diagnostic suggesting renaming one or both of the functions. This feature is useful even without traits so I think documenting it under this title may be a little confusing.

pennae: we've merged it into the name mapping section now and renamed it a bit to make it more general. does that sound good?

hkalbasi: Wouldn't you implement the diagnostic? I won't block this PR on it, but it would help people hitting your problem, maybe saving their day implementing a custom wrapper on their type since they didn't find this feature.

If no, then just add the missing `#` in the subtitle and we are good to go.

hkalbasi: Also a `from` keyword is leftover from your previous design I guess.

pennae: wait what. we deleted the from! git is playing tricks on us. xD

> Wouldn't you implement the diagnostic?

we absolutely would, but we're not sure how to do it. to make the check useful we need position information for each method, but that information is lost early in the transformation process. keeping a mapping of all methods will be very expensive, and duplicating the name mangling procedure into the parser feels like a bad idea. sorry :(

hkalbasi: I see. Ideally `ZngurSpec` should have span data, but that would be a major change, so let's merge this for now.

PR #127, opened 2026-07-11
decrease parser compile times

pennae:
boxing more parsers decreases walltime of zngur-parser compilation by 75% and cpu time by 50%. the parser probably is not greatly sensitive to the performance overhead of boxing, and this makes hacking easier.

### DIFF
--- a/zngur-parser/src/lib.rs
+++ b/zngur-parser/src/lib.rs
@@ -1559,6 +1559,7 @@ fn lexer<'src>()
         .padded()
         .repeated()
         .collect()
+        .boxed()
 }
 
 fn alias<'a>() -> impl Parser<'a, ParserInput<'a>, ParsedItem<'a>, ZngParserExtra<'a>> + Clone {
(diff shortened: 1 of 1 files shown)


hkalbasi:
Thanks for the PR! This is a really important improvement in both dev experience and for users when they clean frequently.

What was your method/heuristic in adding boxed calls? I suspect some of them are not really helping in compile time, and they may even make things worse in some cases.

pennae:
> What was your method/heuristic in adding boxed calls?

"box all return values, choices, and 'large' interior parsers". most functions already return boxed parsers, completing the pattern mostly helps with incremental compilation. boxing choices helps a whole lot in both cases, presumably because we don't need to instantiate as many hidden types in chumsky. boxing large (by looking at lsp type hints) parsers used as arguments to combinators also helps a good bit, but far less than boxing only arguments to choices.

we also tried adding boxes in the middle of long combinator chains, but that didn't have much of an impact. so far we haven't found cases where boxes make compile times worse, in most places they merely don't improve them

hkalbasi:
Seems good!

PR #130, opened 2026-07-24
Fix incorrect reinterpret_cast in field offset calculation.

GoldsteinE:
some missed `const`s, includes a test that would be failing previously

(I’m currently trying to upgrade a real project to a version with variants, so I’ll maybe add more fixes to this branch if I catch anything else)


hkalbasi (approved):
Thanks! Do you want me to hold this to add more fixes into it?

GoldsteinE:
the project passes tests now, so I think that’s the only real fix. we had some discussion about tinkering with signatures of `match` methods to improve ergonomics, and I might try that later, but I think that’s all for now

PR #131, opened 2026-07-24
allow raw strings for cpp_additional_includes

pennae:
otherwise we can't include anything with quotes inside


hkalbasi:
Thanks! Can you please use it in some example too (or add a new example), in order to prevent regressions in future?

pennae:
sure! it's in tutorial_cpp now, hope that's okay

PR #132, opened 2026-07-24
Fix double indirection in field offset calculation.

GoldsteinE:
sorry, part 2 of variant fixes

this was already fixed for primitive types, but not for user-defined types. ideally, we should remove all the code duplication around field offsets, but I don’t yet have courage to attempt such a refactoring.

additionally this:
- removes SCARY comments that marked the place fixing that exact bug for primitives
- changes `examples/enums` to build in debug so that `unreachable_unchecked()` in field offset calculation actually triggers a panic
- uses that to make `examples/enums` test catch this bug

### DIFF
--- a/examples/enums/main.zng
+++ b/examples/enums/main.zng
@@ -21,9 +21,9 @@ mod crate {
         variant First {
             non_exhaustive;
         }
-        // And knows that Second has a 0 field, but not that it's non_exhaustive.
+        // And knows that Second has a 1 field, but not that it's non_exhaustive.
         variant Second {
-            field 0 (offset = auto, type = i32);
+            field 1 (offset = auto, type = NonPrimitive);
         }
 
         fn first() -> Merged;
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,12 @@ jobs:
     runs-on: macos-14
     strategy:
       fail-fast: false
+      matrix:
+        cargo_profile:
+          - debug
+          - release
+    env:
+      EXAMPLES_CARGO_PROFILE: ${{ matrix.cargo_profile }}
     steps:
       - uses: actions/checkout@v3
       - uses: jdx/mise-action@v3
@@ -35,8 +41,12 @@ jobs:
         cpp_compiler:
           - clang++
           - g++
+        cargo_profile:
+          - debug
+          - release
     env:
       CXX: ${{ matrix.cpp_compiler }}
+      EXAMPLES_CARGO_PROFILE: ${{ matrix.cargo_profile }}
     steps:
       - uses: actions/checkout@v3
       - uses: jdx/mise-action@v3
@@ -59,8 +69,13 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
+      matrix:
+        cargo_profile:
+          - debug
+          - release
     env:
       CXX: g++
+      EXAMPLES_CARGO_PROFILE: ${{ matrix.cargo_profile }}
     steps:
       - uses: actions/checkout@v3
       - uses: jdx/mise-action@v3
@@ -75,6 +90,13 @@ jobs:
 
   build-win:
     runs-on: windows-2022
+    strategy:
+      matrix:
+        cargo_profile:
+          - debug
+          - release
+    env:
+      EXAMPLES_CARGO_PROFILE: ${{ matrix.cargo_profile }}
     steps:
       - name: Turn off autocrlf
         run: |
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,8 +22,5 @@ edition = "2024"
 rust-version = "1.85"
 license = "MIT OR Apache-2.0"
 
-[profile.dev]
-opt-level = 3
-
 [profile.release]
 lto = true
--- a/examples/char/Makefile
+++ b/examples/char/Makefile
@@ -1,13 +1,15 @@
-a.out: main.cpp generated.h src/generated.rs src/lib.rs ../../target/release/libexample_char.a
-	${CXX} -std=c++11 -Werror -I. main.cpp -g -L ../../target/release/ -l example_char
+EXAMPLES_CARGO_PROFILE ?= debug
 
-../../target/release/libexample_char.a:
-	cargo build --release
+a.out: main.cpp generated.h src/generated.rs src/lib.rs ../../target/$(EXAMPLES_CARGO_PROFILE)/libexample_char.a
+	${CXX} -std=c++11 -Werror -I. main.cpp -g -L ../../target/$(EXAMPLES_CARGO_PROFILE)/ -l example_char
+
+../../target/$(EXAMPLES_CARGO_PROFILE)/libexample_char.a:
+	cargo build --profile $(EXAMPLES_CARGO_PROFILE:debug=dev)
 
 generated.h ./src/generated.rs: main.zng
 	cd ../../zngur-cli && cargo run g -i ../examples/char/main.zng --crate-name "crate"
 
-.PHONY: ../../target/release/libexample_char.a generated.h clean
+.PHONY: ../../target/$(EXAMPLES_CARGO_PROFILE)/libexample_char.a generated.h clean
 
 clean:
 	rm -f generated.h generated.cpp src/generated.rs a.out actual_output.txt
--- a/examples/char/NMakefile
+++ b/examples/char/NMakefile
@@ -1,13 +1,17 @@
+!IFNDEF EXAMPLES_CARGO_PROFILE
+EXAMPLES_CARGO_PROFILE = debug
+!ENDIF
+
 CXX = cl.exe
 # MSVC doesn't support earlier that c++14
 CXXFLAGS = /W4 /DEBUG /EHsc /std:c++14 /Zc:__cplusplus
-WINLIBS = ntdll.lib
+WINLIBS = ntdll.lib Ws2_32.lib Userenv.lib
 
 EXAMPLE_NAME = char
 
 GENERATED = generated.h src/generated.rs
 
-RUSTLIB_PATH = ../../target/release/
+RUSTLIB_PATH = ../../target/$(EXAMPLES_CARGO_PROFILE)/
 
 RUSTLIB = example_$(EXAMPLE_NAME).lib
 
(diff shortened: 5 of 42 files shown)


hkalbasi:
Thanks! As I said in the telegram group, let's run all examples as debug build in a separate job in CI to catch bugs, instead of selectively having some of them as debug and others as release.

GoldsteinE:
> let's run all examples as debug build in a separate job in CI to catch bugs

let me check if I understand you correctly:
1. all the makefiles (and nmakefiles, ugh) read some env variable (`EXAMPLES_CARGO_PROFILE`?) to set the mode (defaulting to `debug` presumably so local testing is faster)
2. in `.github/workflows/ci.yml` `strategy.matrix` gets a new key to propagate this env var, effectively duplicating existing jobs
3. also probably [these commands in `ci.rs`](https://github.com/hkalbasi/zngur/blob/47a3c49e52cfd65300bad0f1f7de39118f942136/xtask/src/ci.rs#L42) should also read the same var (right now it’s always debug)

hkalbasi:
You are absolutely right :) If we want to not waste microsoft's money, we can check the debug build only on the linux target, but I guess there is no harm in duplicating the jobs.

GoldsteinE:
okay this is harder than I expected for stupid reasons (you need to write `cargo build --profile dev`, but `target/debug`, so you need *two* values to substitute, so you need some logic inside of every makefile to set the second variable based on the first variable), I’ll deal with this tomorrow (probably by just adding a chunk of code into every {n,}makefile, don’t see a good solution there unfortunately)

GoldsteinE:
okay, so, I discovered two things

first of all, debug builds *are* built with optimizations anyway, so I’m not sure how much value there is in doing release builds alongside them. debug assertions might suppress some optimizations I guess, but in these cases we’ll probably hit them?

https://github.com/hkalbasi/zngur/blob/47a3c49e52cfd65300bad0f1f7de39118f942136/Cargo.toml#L25-L26

but also `regression_test1` actually fails in debug currently. I did a little bisecting and the culprit is 64522ea0162c58f78bff1326b07809e7a074ab8c. on this commit, test starts failing with

```
Method call on ZST
[main.cpp:143] zst = 
thread '<unnamed>' (1513872) panicked at examples/regression_test1/src/generated.rs:656:33:
null pointer dereference occurred
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread caused non-unwinding panic. aborting.
zsh: abort (core dumped)  ./a.out
```

<details><summary>(all of the nearby commits also fail with this compile error for me, but after patching it out, the commit I specified is the first one that breaks this test)</summary>

```
error[E0446]: private trait `ZngParser<'src, Self>` in public interface
  --> zngur-parser/src/conditional.rs:15:20
   |
15 |     fn parser() -> impl ZngParser<'src, Self>;
   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak private trait
   |
  ::: zngur-parser/src/lib.rs:73:1
   |
73 | trait ZngParser<'a, Item>: Parser<'a, ParserInput<'a>, Item, ZngParserExtra<'a>> + Clone {}
   | ---------------------------------------------------------------------------------------- `ZngParser<'src, Self>` declared as private

error[E0446]: private trait `ZngParser<'src, Self>` in public interface
  --> zngur-parser/src/conditional.rs:33:20
   |
33 |     fn parser() -> impl ZngParser<'src, Self>;
   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak private trait
   |
  ::: zngur-parser/src/lib.rs:73:1
   |
73 | trait ZngParser<'a, Item>: Parser<'a, ParserInput<'a>, Item, ZngParserExtra<'a>> + Clone {}
   | ---------------------------------------------------------------------------------------- `ZngParser<'src, Self>` declared as private
```

</details>

<details><summary><code>git bisect log</code> for posterity</summary>

```
# bad: [47a3c49e52cfd65300bad0f1f7de39118f942136] allow raw strings for cpp_additional_includes (#131)
# good: [3862e6239c23f342e695d67375c9be688f9d527c] Fix dbg for unsized types and Ref (#21)
git bisect start 'main' '3862e623'
# skip: [a0c63f0f456e8473186fe385464a3cf820800ba4] feat(conditional): functions to build statement parsers
git bisect skip a0c63f0f456e8473186fe385464a3cf820800ba4
# skip: [58508a8831413660f74d9fd1f27daf9b7ff384e0] feat(conditional):  types and parsers for conditional cfg statments
git bisect skip 58508a8831413660f74d9fd1f27daf9b7ff384e0
# skip: [58508a8831413660f74d9fd1f27daf9b7ff384e0] feat(conditional):  types and parsers for conditional cfg statments
git bisect skip 58508a8831413660f74d9fd1f27daf9b7ff384e0
# good: [708a25a331a8d78a836cdec3c6e1f07bd8e3f055] Add noexcept everywhere
git bisect good 708a25a331a8d78a836cdec3c6e1f07bd8e3f055
# good: [708a25a331a8d78a836cdec3c6e1f07bd8e3f055] Add noexcept everywhere
git bisect good 708a25a331a8d78a836cdec3c6e1f07bd8e3f055
# skip: [5f926e3970d634c184d2e7800c07ae42e8da2d80] feat(conditional): pass in cfg values to the cli
git bisect skip 5f926e3970d634c184d2e7800c07ae42e8da2d80
# skip: [17e72bae60f0b705513945557bf830d68d672ad7] Emit separated headers for modules (#109)
git bisect skip 17e72bae60f0b705513945557bf830d68d672ad7
# skip: [aeea4faaee54d60d52106ad603b2295f38edb6e3] feat(conditional): add conditional cfg statements to parser
git bisect skip aeea4faaee54d60d52106ad603b2295f38edb6e3
# good: [469a42f1df0eb722590355dd3d4c7f3838c6d574] Don't use workspace fow wasm example
git bisect good 469a42f1df0eb722590355dd3d4c7f3838c6d574
# bad: [34e5d4aef72425f958b3e746e9cc5bc30e709e6c] Release 0.8.0
git bisect bad 34e5d4aef72425f958b3e746e9cc5bc30e709e6c
# skip: [39c310678d0c05a89f08c71c905cfd296e1344f6] feat(conditional): types to build statement parsers
git bisect skip 39c310678d0c05a89f08c71c905cfd296e1344f6
# good: [8b35a3af663e74be0fb4bc2537b3676314d54906] Improve documentation (#55)
git bisect good 8b35a3af663e74be0fb4bc2537b3676314d54906
# bad: [1c7190140a6baac32b15093215b20382e4a7eea6] Use IndexMap instead of HashMap for determinism (#89)
git bisect bad 1c7190140a6baac32b15093215b20382e4a7eea6
# skip: [34724f6506574d3c006ebf5b5cdadd19049202e1] feat(conditional) add cfg provider trait and impl
git bisect skip 34724f6506574d3c006ebf5b5cdadd19049202e1
# good: [11a6047e7f2a1e4ed0fbf194accc7292f1653e70] Emit C++ with Sailfish templates (#61)
git bisect good 11a6047e7f2a1e4ed0fbf194accc7292f1653e70
# skip: [38328b07c523810f05fdac7aa6a6ec887c2b9bd0] feat(conditional) add parser tests
git bisect skip 38328b07c523810f05fdac7aa6a6ec887c2b9bd0
# good: [6f6cce0eb65ddac3606c29df1902eb032e9e43be] chore: cleanup spelling
git bisect good 6f6cce0eb65ddac3606c29df1902eb032e9e43be
# bad: [64522ea0162c58f78bff1326b07809e7a074ab8c] Support zero-sized types properly.
git bisect bad 64522ea0162c58f78bff1326b07809e7a074ab8c
# good: [c55227b8eba77402624ac3aa5b9c5bd3e414334f] feat(conditional): Make `#if` unstable as well.
git bisect good c55227b8eba77402624ac3aa5b9c5bd3e414334f
# first bad commit: [64522ea0162c58f78bff1326b07809e7a074ab8c] Support zero-sized types properly.
```

</details>

I’m a bit unsure on the next steps. I’d really like to get the fix in this PR in, but testing it properly requires running tests in debug mode, and doing that won’t pass CI because of an unrelated bug. I’ll take a look at it and see if I can fix it quickly, but I’m not sure I’d be able to do that

GoldsteinE:
wow that’s wild.

so the issue is: ZSTs use `std::array<uint8_t, 0>` and for some reason I can not possibly comprehend calling `.data()` on such an array “may or may not” return a null pointer. it *does* return a null pointer, which then gets converted into a rust reference, which is UB. they should probably use, like, something else, or maybe take a pointer *to* the array instead of calling `.data()`, or something. idk. I need to process this information.

GoldsteinE:
okay, I fixed the bug, let’s try this out

GoldsteinE:
what does any of that even mean... linker errors seem to refer to, like, stdlib functions? related to sockets? what?
did release build eliminate dead code better so these symbols weren’t needed but now we need to link to smth else?

GoldsteinE:
I think gh actions is tired of me lol

https://github.com/hkalbasi/zngur/actions/runs/30157952748

hkalbasi:
Seems like we opened a can of worms.

> opt-level = 3

Git blame says it is from initial commit. Interesting. I think we should remove it and can't find a valid reason for optimizing debug builds in zngur.

> so the issue is: ZSTs use std::array<uint8_t, 0> and for some reason I can not possibly comprehend calling .data() on such an array “may or may not” return a null pointer.

So what did you do? I guess we may need to emit an array with size 1 (which is not much different perf wise since there is no real zero-sized type in C++).

hkalbasi:
> I think gh actions is tired of me lol

It is down for everyone
https://www.githubstatus.com/

GoldsteinE:
> So what did you do?

instead of calling `.data()` I just `reinterpret_cast` the pointer to the array itself, which should be just as valid and doesn’t create the weird edgecase around ZSTs. here’s the fix commit: https://github.com/hkalbasi/zngur/pull/132/changes/1346669d2aa941f379304cf8725d1f5320f1d9bc

> Seems like we opened a can of worms.

both fixes & CI changes are basically ready, I’m just waiting for github to stop having its second outage in two days so I can fixup windows CI (which now needs a few more `WINLIBS` due to less DCE or smth)

> Git blame says it is from initial commit. Interesting. I think we should remove it and can't find a valid reason for optimizing debug builds in zngur.

sure, I’ll remove it

GoldsteinE:
now it’s ready (complete with a nice commit history, no need to squash lol)
fixes both bugs, disables `opt-level = 3` for dev builds and runs every CI job twice

examples/enums/main.zng
```diff
@@ -21,9 +21,9 @@ mod crate {
         variant First {
             non_exhaustive;
         }
-        // And knows that Second has a 0 field, but not that it's non_exhaustive.
+        // And knows that Second has a 1 field, but not that it's non_exhaustive.
         variant Second {
-            field 0 (offset = auto, type = i32);
+            field 1 (offset = auto, type = NonPrimitive);
```
hkalbasi, line 26: Shouldn't this get a compile error?

GoldsteinE: why? it’s marked as `non_exhaustive` in another file, so after merging both variants are `non_exhaustive`

https://github.com/hkalbasi/zngur/blob/d66fc0d6c245b92b1ea37dc8e5424c74a184cf06/examples/enums/second.zng#L16-L18

this is a test that exercises as much of merging logic as possible

hkalbasi: Ah, I see.

hkalbasi:
Thanks!

PR #137, opened 2026-08-25
Restore `operator T()` on `Field*` and allow `operator(*|->)` on `Ref/RefMut`

### LINKED ISSUE #133: `Copy` type field access is blocked
`FieldOwned<T>::operator T()` is defined on the primary template but per-type explicit specializations drop it. Types that have a specialization lose the conversion operator.

Built-in primitives like `int32_t` work because they have no specialization. `Bool` doesn't work.

My (naive?) proposal would be a [CRTP](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern) mixin that provides `operator T()` for all field kinds and primary templates / specializations should inherit from it. If you are OK with this, I can contribute it.

RaitoBezarius:
Fixes #133.

Each commit presents its rationale.

TL;DR:

- Add `operator*` and `operator->` (const/non-const) for `Ref/RefMut`
- Add `operator T()` on `FieldRef/FieldMut/FieldOwned` using `operator*` on `Ref`/`RefMut`
- Added some regression testing for my usecases

### DIFF
--- a/examples/regression_test1/expected_output.txt
+++ b/examples/regression_test1/expected_output.txt
@@ -59,19 +59,35 @@ Test Field* underlying conversions -- started
 [main.cpp:88] pmut.f1.len() = 3
 Test Field* underlying conversions -- finished
 
+Test Ref and RefMut deref of user types -- started
+[main.cpp:101] view.flag = true
+[main.cpp:102] int32_t(view.data.buzz_3) = 9
+[main.cpp:107] snapshot = BoolHolder {
+    flag: true,
+    data: FieldTypeC {
+        buzz_1: 7,
+        buzz_2: 8,
+        buzz_3: 9,
+    },
+}
+[main.cpp:110] int32_t(r->data.buzz_3) = 9
+[main.cpp:115] holder.flag = false
+[main.cpp:117] holder.flag = true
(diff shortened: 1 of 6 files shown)


hkalbasi:
Thanks for the PR! I think adding a general propose `T& operator*()` is impossible due to layout deferences between C++ type and Rust type. What is possible is a `T operator*()` for `Copy` types (and the current implementation is UB for anything non trivial). Similarly, we can't have a normal `->` operator and what we actually do is repeating all methods and fields of the `T` on `Ref<T>` too (which also helps in having `.` instead of `->` which is more like Rust and C++ references). The field `Copy` implementation looks good to me.

RaitoBezarius:
> Thanks for the PR! I think adding a general propose `T& operator*()` is impossible due to layout deferences between C++ type and Rust type. What is possible is a `T operator*()` for `Copy` types (and the current implementation is UB for anything non trivial). Similarly, we can't have a normal `->` operator and what we actually do is repeating all methods and fields of the `T` on `Ref<T>` too (which also helps in having `.` instead of `->` which is more like Rust and C++ references). The field `Copy` implementation looks good to me.

I will drop the implementations and rework it accordingly, thanks.

PR #147, opened 2026-09-17
better const str concat

Ryex:
I've had this sitting around for two months and forgot about it

improve the `const str` concat code
  - use `::core` for better hygiene
  - use better algorithm for more concise code, and hopefully better compile times (however marginal) 
    -  (don't convert each `str` into an array first, just computer total len + no more recursive macro)


hkalbasi:
`let ::core::ops::Range { mut start, end } = 0..strs.len();` is a fun way to define two variables start and end :)

Ryex:
> `let ::core::ops::Range { mut start, end } = 0..strs.len();` is a fun way to define two variables start and end :)

welcome to const rust :)

Issue #142, opened 2026-08-31
Feature request: avoid manual 'pub use generated::cpp::X;' re-export for C++ types


camio:
Is there a way to make this automatic? If I understand correctly, this is the only `.rs` boilerplate required when exposing C++ to Rust. Perhaps there's a good reason for requiring the "pub use" in `lib.rs` that I'm missing?

hkalbasi:
It replaced the previous boilerplate `pub struct X(ZngurOwnedObject)` in the lib.rs file. Do you have a better suggestion? It is needed since we want to generate Rust code only in the `generated` module, so we can't add items to the root directly.

Ryex:
if we had a way to inform the generator what module the generated code would live under perhaps?

hkalbasi:
It already knows (the expected path is in the zng file) but I don't think zngur (or any other build process) should be able to change anything other than build artifacts?

camio:
C++ functions, on the Rust side, are always put in `::self`. There is no way to specify a different module to put those in.

C++ types, on the Rust side, are always placed in `::self::cpp`. You can specify a different module, but it is always ignored as far as Rust is concerned. On the C++ side, these Rust wrapper types are placed in a `rust::` namespace, followed by the specified module name.

What if we *always* put C++ types in `::self` instead of `::self::cpp` and disallow putting them elsewhere? This would be consistent with functions and eliminate the need for `pub use generated::cpp::X;`.

We can refer to these types elsewhere in the .zng specification using the `self::` prefix. When generating the C++ code, we can substitute `::rust::` for `self::`. I think that would work.

hkalbasi:
The difference between functions and types is that types also have a Rust path (to support impl blocks for them and to use them in other parts of the zng file) unlike functions which live in a `extern "C++"` block. I don't think the problem here is the `::cpp` submodule, removing it just changes `pub use generated::cpp::X;` to `pub use generated::X;`.

hkalbasi:
We can have some kind of `<extern "C++">::X` path in the zng file to mention C++ types that have no canonical Rust path in the zng file, and those types could dodge the `pub use generated::X`.

camio:
Consider this `.zng` file (See https://github.com/camio/zngur-use-experiment for the full example):

```
#cpp_additional_includes r#"
    #include "foo.h"
"#

type ::foo::Foo {
    #layout(size = 16, align = 8);
    #cpp_heap_allocated "::foo::Foo";
}

extern "C++" {
[...]

PR #148, opened 2026-09-18
Add c++:: path syntax for C++ types with no canonical Rust path (#142)

camio:
This is an implementation of #142 following the consensus design there.

Paths in the .zng file may be prefixed with `c++::` to refer to the Rust wrapper of a C++ type. It can be used in `mod` blocks, type names, and anywhere you might need a type.

We also put all generated types in the top-level namespace of `generated.rs` or wherever you put your generated type. Type aliases are placed in `cpp::` for backward compatibility, and this is done as a deprecated alias to encourage migration.

The tutorial and all examples were updated to make use of this new feature.

A lot of the code was simplifying what was there, which was nice. The most difficult part was getting it working properly with `use` and templates.

The following text is from the AI.

- David

## Summary
- Adds `c++::a::b::Name` path syntax for declaring/referencing a C++ wrapper type with no canonical Rust path, and `mod c++::a::b { ... }` as a scoping shorthand for grouping several such types.
- Removes the need for the `pub use generated::cpp::X;` boilerplate this issue is about: new-style types are placed directly at the top level (or in a nested module mirroring their `c++::` segments) of the generated code, with no `generated::cpp` involved at all.
- `generated::cpp` is kept as a `#[deprecated]` compatibility alias for old-style declarations only, so existing code keeps compiling with a deprecation warning.
- Migrates the `stack_owned`, `tutorial_cpp`, and `regression_test2` examples to the new syntax; `multiple_modules/aggregation` stays old-style since its type is referenced cross-crate by Rust path, which `c++::` types can't support.
- Updates `opaque.md` and `tutorial.md` to match.
- Along the way, fixes two related correctness bugs uncovered during review (a free `fn` silently mis-resolving when declared inside a `c++::` scope, and a `c++::`-only path silently mis-resolving when reached via alias indirection into trait position) and renames `RustType::CppOnly` to `RustType::Cpp`.

## Test plan
- [x] `cargo test` passes across zngur-parser (60 tests), zngur-def, zngur-generator, and zngur under `RUSTFLAGS="-D warnings"`.
- [x] `cargo check`/`cargo fmt --check` clean across all core crates.
- [x] `cspell .` and the book's `dprint` formatting check both pass clean.
- [x] All 24 buildable examples (everything except `tutorial-wasm32`) rebuilt and diffed against their `expected_output.txt`; all match.


camio:
I think a good follow-up to this PR would be to remove the necessity of specifying C++ paths a second time from within a type block.

```
type c++::Way {
    #cpp_ref "::osmium::Way";
}
```

would become

```
type c++::osmium::Way {
    #cpp_ref;
}
```

, or alternatively,

```
mod c++::osmium {
  type Way {
      #cpp_ref;
  }
}
```

zngur-parser/src/tests.rs
```diff
@@ -1202,3 +1204,498 @@ fn cpp_additional_includes() {
...
+    }
+}
+    "#,
+        NullCfg,
+        |_| {},
+    );
+    let direct_ty = &direct.spec.types.first().expect("no type parsed").ty;
+    let via_mod_ty = &via_mod.spec.types.first().expect("no type parsed").ty;
+    assert_cpp(&["a", "b", "Name"], direct_ty);
+    assert_eq!(direct_ty, via_mod_ty);
```
Ryex, line 1306: I think the ai got a bit confused with this test.
`direct_ty` and `via_mod_ty` are tautologically the same reference

zngur-parser/src/lib.rs
```diff
@@ -951,7 +1032,12 @@ struct ParsedRustPathAndGenerics<'a> {
 impl ParsedRustPathAndGenerics<'_> {
     fn to_zngur(self, scope: &Scope<'_>) -> RustPathAndGenerics {
         RustPathAndGenerics {
-            path: scope.resolve_path(self.path),
+            path: match scope.resolve_path(self.path) {
+                EntityPath::Rust(v) => v,
+                EntityPath::Cpp(_) => todo!(
+                    "a c++::-only path was used somewhere that can't support it (e.g. as a trait); needs a proper user-facing error instead of this panic"
+                ),
```
Ryex, line 1039: regarding this and other errors generated when converting `Parsed*` into into spec items.

Not something to address in this Pr but we should probably follow `rustc_public`'s lead and save a reference to our source map and parse context into a [scoped thread_local](https://docs.rs/scoped_thread_local/latest/scoped_thread_local/) so that we can emit proper errors with scopes

Ryex (commented):
this looks good.

hkalbasi:
I have a more ambitious plan for `c++::Ty` types. I think we had too many iterations on this feature, and its time to do it for the last time, correctly.

* Types in the `c++::` namespace are normal types:
  * They use normal layout policies (`#only_by_ref` instead of `#cpp_ref`, `#heap_allocated`, `#layout_conservative`, ...)
  * If it uses `#layout` or `#layout_conservative`, it needs either `Copy` (trivially copiable), `Move` (trivially relocatable) or `!Move` (`!Move` needs future Rust support, we won't use `Pin` trick ever)
* `c++::` types can accept template, e.g. `c++::std::vector<crate::RustType>` or `c++::std::unique_ptr<c++::CppType>`. I have a proof that this is possible but it is too large to fit the margin.
* Zngur and Rust type bijection to C++ Type:
    * `&c++::Ty` -> `Ref<Ty>` and `&mut c++::Ty` -> `RefMut<Ty>` (Zngur will emit `Ref`/`RefMut` specializations for all `c++::` types with a `.cpp()` method, similar to now)
    * `*const c++::Ty` -> `const Ty*` and `*mut c++::Ty` -> `Ty*` (No `Raw`/`RawMut` specialization is emitted)
    * There is no Rust type for C++ references, so C++ functions/methods working with those need glue code (which is needed for other semantics alignments anyway)
* Users can provide `impl` and `impl Trait` for `c++::` types in C++, e.g. `rust::Impl<std::vector<int32_t>, rust::Iterator<int32_t>>`.

Obviously, all of this doesn't need to happen in a single PR, but whatever part that is implemented should follow (a more discussed and polished future version of) this. I didn't thoroughly looked at this implementation, it might be close, but it diverges at least about the layout policy keywords (it uses the old ones which I think are inferior).

Opinions?

camio:
> I have a more ambitious plan for `c++::Ty` types. I think we had too many iterations on this feature, and its time to do it for the last time, correctly.
> 
> ...
> 
> Obviously, all of this doesn't need to happen in a single PR, but whatever part that is implemented should follow (a more discussed and polished future version of) this. I didn't thoroughly looked at this implementation, it might be close, but it diverges at least about the layout policy keywords (it uses the old ones which I think are inferior).
> 
> Opinions?

I think this design is brilliant. The unification of types makes .zng files so much simpler!

This PR, as it stands, serves as an incremental stepping stone towards that vision. It removes the "cpp" module, introduces "c++", and eliminates the need for "use" on types. After this PR lands, I'd be happy to help contribute more progress towards your ultimate vision. Good next steps would be to support the normal layout policies for C++ types and to use the type's zng module path to figure out the fully qualified C++ type.

PS. Love the Fermat reference!

zngur-parser/src/lib.rs
camio: Oh, interesting. I was going to thread it through, but then though this PR might be too difficult to review if I did that.

zngur-parser/src/tests.rs
camio: Good catch. I didn't spend much time looking at the tests, so that, of course, is where AI does its worst. LOL.

I'll fix this tomorrow.

zngur-parser/src/lib.rs
Ryex: threading it through is a valid option too. 
just finished up a RP to improve the delivery and generation of reports so that might help you with ideas here.

zngur-parser/src/tests.rs
camio: I took a closer look at this, and I think the test is actually correct. direct_ty is coming from the "type c++::a::b::Name" snippet and via_mod_ty is coming from the "mod c++::a::b { type Name" snippet. These two lines verify that both of them parse to the same `["a", "b", "Name"]` path. Make sense?

Ryex: I suppose that's a valid test of name resolution.

hkalbasi:
> I think this design is brilliant. The unification of types makes .zng files so much simpler!

Thanks for the kind words! I hope this time we can settle on a design for a long time.

> This PR, as it stands, serves as an incremental stepping stone towards that vision.

I think it is generally in the right direction, but it diverges in some details. E.g. I don't want to use `#cpp_ref` here and then immediately deprecate/remove it and use `#only_by_ref` in the next PR. We can do breaking changes to fix our mistakes, but I don't want to create churn. I'm in favor of doing incremental and small changes, e.g. we can make `#layout_conservative` a not supported error for `c++::` types, but I guess using `#only_by_ref` and `#layout` instead of `#cpp_ref` and `#cpp_stack_allocated` wouldn't make this PR much more complex (it might even reduce some complexity by code reuse between them and normal type).

So I think we should 1. agree on a design, fix its problems and potential improvements (it isn't set in stone, we can make breaking change later with more feedback / details from implementation, but I think it is good to look more than the immediate step and have a vision on the future plan) 2. make this PR a forward compatible subset of that design.

camio:
> I think it is generally in the right direction, but it diverges in some details. E.g. I don't want to use `#cpp_ref` here and then immediately deprecate/remove it and use `#only_by_ref` in the next PR. We can do breaking changes to fix our mistakes, but I don't want to create churn.

Your churn concerns are valid. Moreover, deprecation definitely creates technical debt. In this specific case, however, I think the tradeoff is worth it to allow for incremental migrations. If a user can switch to `c++::` and remove `use` in one step (commit/test/push/release), then switch to `#only_by_ref` and `#layout` in another step, it significantly reduces the friction of adopting the new best practices.

> So I think we should 1. agree on a design, fix its problems and potential improvements (it isn't set in stone, we can make breaking change later with more feedback / details from implementation, but I think it is good to look more than the immediate step and have a vision on the future plan) 2. make this PR a forward compatible subset of that design.

I'm fully on board with first outlining the larger design vision so we know this PR is forward-compatible. Would you have the bandwidth to take a first pass at documenting that design? I’d like to keep the momentum going on this PR to get its concrete benefits into users' hands.

hkalbasi:
> In this specific case, however, I think the tradeoff is worth it to allow for incremental migrations.

IME the number of breaking changes is as important as the volume of change, it would be annoying to touch everything twice. But if you think having `#cpp_ref` and similar helps in migration, I have no objection on accepting them *in addition to* normal layouts, so that both one time and incremental migrations become possible.

PR #149, opened 2026-09-19
ZngSpec Printing via cargo feature

Ryex:
adds a printing feature to the `zngur-def` crate to print the zng IDL from a spec (and or contained types)
what parts of the IDL get printed is controllable.

this will be what does the final output from autozng


Ryex:
I'm going to follow this up with better error handling in the parser and then some reparse intigration tests for this feature.

